PR #4, opened 2022-03-12
feat/beam

rvilarl:
I am having problems with ESM modules.
`TypeError: _vexflow.Factory is not a constructor`

@jaredjj3 any idea?


jaredjj3:
I'm not sure, can you just use the named exports for now?

rvilarl:
@ronyeh can you help us?

PR #11, opened 2022-04-05
measure start & end messages

rvilarl:
I have added passacaglia.xml. It is only printing the first two measures fixed width.

image

### DIFF
--- a/src/producer.ts
+++ b/src/producer.ts
@@ -29,16 +29,16 @@ export class Producer {
     const nodeElem = node as Element;
     switch (node.nodeName) {
       case 'measure':
-        const width = nodeElem.getAttribute('width') ?? '';
-        const staves = nodeElem.getElementsByTagName('staves').item(0)?.textContent ?? '1';
+        const width = parseFloat(nodeElem.getAttribute('width') ?? '300');
+        const staves = parseFloat(nodeElem.getElementsByTagName('staves').item(0)?.textContent ?? '1');
         const sign: string[] = [];
-        for (let i = 0; i < parseInt(staves); i++) {
+        for (let i = 0; i < staves; i++) {
           sign.push(nodeElem.getElementsByTagName('sign').item(i)?.textContent ?? '');
         }
         const beats = nodeElem.getElementsByTagName('beats').item(0)?.textContent ?? '';
         const beatType = nodeElem.getElementsByTagName('beat-type').item(0)?.textContent ?? '';
-        messages.push({ msgType: 'measureStart', width, staves, clef: sign, time: `${beats}/${beatType}` });
-        messages.push(...Array.from(node.childNodes).flatMap(this.getMessages));
+        messages.push({ msgType: 'measureStart', width, staves, clefs: sign, time: `${beats}/${beatType}` });
+        messages.push(...Array.from(node.childNodes).flatMap(this.getMessages.bind(this)));
         messages.push({ msgType: 'measureEnd' });
         break;
       case 'attributes':
@@ -70,7 +70,7 @@ export class Producer {
         if (beam === 'end') messages.push({ msgType: 'beamEnd' });
         break;
       default:
-        console.log(`unprocessed node, got:`, node);
+      // console.log(`unprocessed node, got:`, node);
     }
     return messages;
   }
--- a/src/producer.ts
+++ b/src/producer.ts
@@ -14,7 +14,6 @@ export class Producer {
 
   message(receiver: EasyScoreMessageReceiver): void {
     const cursor = Cursor.fromMusicXml(this.musicXml);
-    receiver.onMessage({ type: 'voiceStart' });
     while (cursor.hasNext()) {
       const node = cursor.next();
       const messages: EasyScoreMessage[] = [];
@@ -23,7 +22,6 @@ export class Producer {
         receiver.onMessage(message);
       }
     }
-    receiver.onMessage({ type: 'voiceEnd' });
   }
 
   private getMessages(node: Node): EasyScoreMessage[] {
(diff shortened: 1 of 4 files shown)


src/types.ts
```diff
@@ -1,44 +1,39 @@
-export type VoiceStartMessage = {
-  type: 'voiceStart';
+export type MeasureStartMessage = {
+  msgType: 'measureStart';
+  width: string;
+  staves: string;
```
jaredjj3, line 4: The `<staves>` element has a `nonNegativeInteger` for its data type. https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/staves/

Would you change this type to a `number` please?

src/types.ts
```diff
@@ -1,44 +1,39 @@
-export type VoiceStartMessage = {
-  type: 'voiceStart';
+export type MeasureStartMessage = {
+  msgType: 'measureStart';
+  width: string;
```
jaredjj3, line 3: I'm not a fan of using empty string `''` to imply a certain behavior. We need a more advanced type here to allow callers to explicitly not specify a width. For example:

```ts
type Width = {
  widthType: 'unspecified';
} | {
  widthType: 'specified';
  value: number;
  units: 'px' | 'cm' | 'in'; // etc
};
```

src/types.ts
```diff
@@ -1,44 +1,39 @@
...
 
 export type BeamEndMessage = {
-  type: 'beamEnd';
+  msgType: 'beamEnd';
 };
 
 export type NoteMessage = {
-  type: 'note';
+  msgType: 'note';
+  rest: boolean;
```
jaredjj3, line 23: I know that EasyScore treats rests as notes, but we should just create a separate `RestMessage`. Most of the properties in `NoteMessage` are not applicable to rests.

src/types.ts
```diff
@@ -1,44 +1,39 @@
-export type VoiceStartMessage = {
-  type: 'voiceStart';
+export type MeasureStartMessage = {
+  msgType: 'measureStart';
+  width: string;
+  staves: string;
+  clef: string[];
```
jaredjj3, line 5: nit: Since there are multiple, can we call this `clefs` instead?

src/types.ts
```diff
@@ -1,44 +1,39 @@
...
+  msgType: 'beamEnd';
 };
 
 export type NoteMessage = {
-  type: 'note';
+  msgType: 'note';
+  rest: boolean;
   stem: string;
   pitch: string;
+  type: string;
```
jaredjj3, line 26: In my opinion, `type` is too ambiguous for anything other than a discriminant. Can we call this `graphicNoteType` instead please? https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/type/

src/producer.ts
```diff
@@ -23,53 +22,55 @@ export class Producer {
...
+        const width = nodeElem.getAttribute('width') ?? '';
+        const staves = nodeElem.getElementsByTagName('staves').item(0)?.textContent ?? '1';
+        const sign: string[] = [];
+        for (let i = 0; i < parseInt(staves); i++) {
+          sign.push(nodeElem.getElementsByTagName('sign').item(i)?.textContent ?? '');
         }
+        const beats = nodeElem.getElementsByTagName('beats').item(0)?.textContent ?? '';
+        const beatType = nodeElem.getElementsByTagName('beat-type').item(0)?.textContent ?? '';
+        messages.push({ msgType: 'measureStart', width, staves, clef: sign, time: `${beats}/${beatType}` });
+        messages.push(...Array.from(node.childNodes).flatMap(this.getMessages));
```
jaredjj3, line 41: nit: This was my fault, I think it needs to be `this.getMessages.bind(this)` to not lose context.

src/producer.ts
```diff
@@ -23,53 +22,55 @@ export class Producer {
...
         });
         const beam = '' + nodeElem.getElementsByTagName('beam').item(0)?.textContent;
-        if (beam === 'begin') messages.push({ type: 'beamStart' });
-        if (beam === 'end') messages.push({ type: 'beamEnd' });
+        if (beam === 'begin') messages.push({ msgType: 'beamStart' });
+        if (beam === 'end') messages.push({ msgType: 'beamEnd' });
         break;
       default:
-        console.log(`unprocessed node, got: ${node}`);
+        console.log(`unprocessed node, got:`, node);
```
jaredjj3, line 73: We don't have to do this now in this PR, but understanding what nodes we're skipping is crucial to understanding what we're not accounting for. We should eventually try to format this to have line numbers, maybe show a snippet of the XML, etc. to help us in the future.

src/renderer.ts
```diff
@@ -1,6 +1,6 @@
 import * as VF from 'vexflow';
 import { Producer } from './producer';
-import { ClefMessage, EasyScoreMessage, NoteMessage } from './types';
+import { EasyScoreMessage, MeasureStartMessage, NoteMessage } from './types';
 
 export class Renderer {
   static render(elementId: string, musicXml: string): void {
```
jaredjj3, line 6: When I make the width and height of the `VF.Factory` renderer large, I can see the entire score:

image

Is there a way to auto-size the renderer?

src/renderer.ts
```diff
@@ -26,54 +26,89 @@ export class Renderer {
...
+              notes: notes.get(curStaff)!.get(curVoice)!.slice(beamStart),
+              options: { autoStem: true },
+            });
             beamStart = -1;
           }
           break;
-        case 'voiceStart':
-          notes = [];
+        case 'measureStart':
+          system = this.factory.System({ x, width: 300 });
```
jaredjj3, line 56: Shouldn't we just use the width from the message?

jaredjj3 (changes requested):
Thanks for adding this feature! I'm glad that you're able to build on top of the current abstractions.

I know I added a lot of comments, but we can just keep reviews to one round if you wish to keep momentum going.

src/producer.ts
jaredjj3: Or even better, surface this information so that we can render it in the dev app. I can take this on.

src/types.ts
rvilarl: I suggest to leave it as defined by MusicXML and this is a number. I change it to number.

src/types.ts
rvilarl: done

src/types.ts
rvilarl: done

src/types.ts
rvilarl: As said before I would try to keep MusicXML approach and there, rest is an attribute of note.

src/types.ts
rvilarl: type is the MusicXML naming.

src/renderer.ts
rvilarl: I will take that on the next PR

src/producer.ts
rvilarl: I leave that to you. I cooment out for now.

src/producer.ts
rvilarl: Done

jaredjj3:
RE: The last two remaining comments. Feel free to merge what you have, but I think we need to make an important design decision going forward:

1. **Try to adhere to the MusicXML structure as much as possible.** The pros are that MusicXML is a battle-tested spec and there is evidence that we can render most musical notations with it (i.e. we can import a MusicXML file into a lot of popular software and it'll work). The con is that it weakens the `EasyScoreMessage` abstraction. It's trivial to just convert MusicXML to JSON (or some other queryable data structure like an `ElementNode`), and then just pass that to the `vexml.Renderer`. Additionally, it's not great that a `vexml.MessageReceiver` will have to know what to expect in the MusicXML spec, which is very large and has a lot of branches.

2. **Create our own abstractions if it makes message producing and/or receiving easier.** The pros is that we have a mechanism to incrementally collect score information (i.e. the `*Start` and `*End` messages). `vexml.MessageReceiver`s doesn't have to implement a complex "lookahead" mechanism that searches for needed context. Instead they can wait for a message that has it. Another pro is that we can define interfaces that makes it easier to work with `vexflow.EasyScore`. However, this is all at the expense of us misjudging the abstraction boundaries, and having to go back and redefine messages as we add features - just like how you did in this PR.

3. **Some combination of (1) and (2), or something else ???**

My vote is for (2) because one of the motivations of creating the `EasyScoreMessage` abstraction was to make it easy to work with `EasyScore`, which is think is more important than encoding music, which is what MusicXML was made for. After all, we don't have to worry about going from `EasyScoreMessage` -> MusicXML. Managing MusicXML documents is a separate concern outside of this library.

jaredjj3 (approved):

PR #13, opened 2022-04-06
Feat/codeprint

rvilarl:
`codePrint` feature. Looking at the output I would say that I should refactor to avoid the double indented map. :)


jaredjj3:
Now I understand what you meant by this feature. It seems like you only care about what is being passed to the `vexflow` objects. I think I have an idea involving a [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) that will do the same thing. I’ll need some time to iron it out.

Also, yeah the nested maps are really confusing. If I was a prospective Vexflow user, I would be intimidated with how much I’d have to write to produce a small output.

PR #14, opened 2022-04-06
refactor: removal of the Map

rvilarl:
What do you think?

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -30,8 +30,8 @@ export class Renderer {
 
     let timeSignature = '';
     let clefs: string[] = [];
-    // Map<staff,Map<voice,Notes[]>
-    const notes = new Map<string, Map<string, VF.StemmableNote[]>>([]);
+    let voices: VF.Voice[] = [];
+    let notes: VF.StemmableNote[] = [];
     let beamStart = -1;
     let curVoice = '0';
     let curStaff = '0';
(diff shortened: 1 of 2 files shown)


src/renderer.ts
```diff
@@ -74,36 +75,35 @@ export class Renderer {
           if (message.accidental != '') {
             note.addModifier(this.factory.Accidental({ type: this.getAccidental(message.accidental) }));
           }
-          // New Staff, add previous to system (TODO function required)
+          // New Staff, add previous to system
```
rvilarl, line 78: This could be a StaffEndMessage

src/renderer.ts
```diff
@@ -74,36 +75,35 @@ export class Renderer {
...
-            });
+            voices.push(score.voice(notes));
+            notes = [];
             if (system) {
               const stave = system.addStave({ voices: voices });
               if (this.getClef(clefs, curStaff) !== '') stave.addClef(this.getClef(clefs, curStaff));
               if (timeSignature !== '/') stave.addTimeSignature(timeSignature);
             }
+            voices = [];
+          } else if (curVoice !== '0' && curVoice !== message.voice) {
```
rvilarl, line 88: This could be a VoiceEndMessage

rvilarl:
I will merge this one, the code is simpler. I will also update #13 accordingly to assess #16

PR #16, opened 2022-04-07
CodePrinter

jaredjj3:
@rvilarl, I built this off of your idea in  https://github.com/stringsync/vexml/pull/13. This PR is what I meant by using `Proxy` to trap all the getter/setter calls and generate the code without rewriting it.

Currently, all `codeprinter` tests pass, but they are far from exhaustive. I still need to:

- add some functionality that declares variables
- accounted for nested getters/setters.

I would like to use the output of this `CodePrinter` to show in the dev server app since it may help immensely with debugging.

I didn't finish, but I wanted to upload what I had to get your feedback.


rvilarl:
@jaredjj3 my only concern is that it might end up being more complex than just adding the codePrint statements. Anyway I am exited to see how it integrates and work. :)

jaredjj3:
Haha, it is absolutely overengineered, but I saw this as a fun opportunity :)

I think this will help gain support from people who work on Vexflow because it will be very obvious what we're trying to do. DX is very important to me, so I wanted to make sure you can continue to print code without having to write it twice.

jaredjj3:
@rvilarl I'm excited to share that I was able to get it to work! All the tests in src/codeprinter.test.ts pass, which have all the examples of how the `CodePrinter` is used. As you can see, the unit tests now assert that the watch expressions are automatically declared.

The next steps are wiring it into `vexml.Renderer`, observing what breaks, adding tests, and then hopefully being ready for review in a few days. I'll consider abandoning this complex solution if it takes more than a week.

rvilarl:
> The next steps are wiring it into `vexml.Renderer`, observing what breaks, adding tests, and then hopefully being ready for review in a few days. I'll consider abandoning this complex solution if it takes more than a week.

Looking forward to seeing the results of the first wirings!

rvilarl:
I will make an uüpdate to master before you wire to support chords, fix notes in bass staves and use measure.width.

jaredjj3:
Sounds good, I'll focus on tests and wait for your changes on `master` before I wire anything in. I can also work on the dev server while I wait. Let me know when you finish.

rvilarl:
done, I had it prepared

jaredjj3:
Ack, thanks for the update.

jaredjj3:
@rvilarl, it's ready for review. Feel free to merge if you're happy with the results. After this is merged, you might need to install dependencies by running `yarn` in the root dir.

https://user-images.githubusercontent.com/19232300/162630323-7789b720-dabc-43d4-832e-71f6e993606a.mov

rvilarl (commented):
It is indeed overengineer but it looks very good! I like it.

jaredjj3:
Happy to hear! I thought I could get rid of all the duplication, but there were some edge cases that I couldn't fully figure out, which is why there are `CodePrinter.literal` calls sprinkled around. `CodePrinter.expression` works for most cases and it does allow you to only have to write an expression once, so I think it's a t least a partial win. Feel free to convert to your original codePrint idea if it gets in the way too much.

PR #25, opened 2022-04-12
grace support included

rvilarl:
I had to move away from `easyscore` to `factory`. This was already suggested by Mohit in VexFlow that it makes more sense to use this interface for the library.


jaredjj3:
Good call. EasyScore is meant to be consumed by humans and it's naturally not as capable as the lower level structures.

rvilarl:
@jaredjj3 here are all my fixes so far, I also had to move away from System as it does not allow voices to jump between staves. A lot of lilypond is runnig although many features are missing.

src/types.ts
```diff
@@ -1,9 +1,14 @@
...
   msgType: 'measureStart';
-  width: number;
-  staves: number;
-  clefs: string[];
-  time: string;
+  width?: number;
+  staves?: number;
+};
+
+export type AttributesMessage = {
```
jaredjj3, line 7: nit: This is a very generic name. Would you add a comment referencing https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/attributes/ please?

src/types.ts
```diff
@@ -1,9 +1,14 @@
...
+  msgType: 'attributes';
+  clefs: Map<number, string>;
```
jaredjj3, line 9: Later on, I'd like to stick to types that are easily serializable to make it easy to send these messages over-the-wire (think multiple people editing the same MusicXML). No need to change now, but later on, I'd like this to be a `Record<number, string>`.

src/renderer.ts
```diff
@@ -38,27 +43,31 @@ export class Renderer {
...
-    let system = t.let<VF.System | undefined>('system', () => undefined);
     let timeSignature = t.let('timeSignature', () => '');
-    let clefs = t.let('clefs', () => new Array<string>());
     let clef = t.let('clef', () => '');
-    let curClefs: string[] = [];
-    let voices = t.let('voices', () => new Array<VF.Voice>());
-    let note = t.let<VF.StemmableNote | undefined>('note', () => undefined);
-    let notes = t.let('notes', () => new Array<VF.StemmableNote>());
+    let stave: VF.Stave | undefined = t.let('stave', () => undefined);
+    const staves: Map<number, VF.Stave> = t.let('staves', () => new Map<number, VF.Stave>());
```
jaredjj3, line 51: nit: Here and below, use `t.const` instead so it prints correctly.

src/renderer.ts
```diff
@@ -68,70 +77,101 @@ export class Renderer {
...
+              noteStruct.keys.push(`${head.pitch}`);
             }
-            name = name.replace(/ /, '(');
-            name += `)/${durationDenominator}`;
+            noteStruct.duration = `${durationDenominator}`;
           }
 
+          if (message.grace) {
+            noteStruct.slash = message.graceSlash;
+            note = this.factory.GraceNote(noteStruct);
```
jaredjj3, line 166: nit: To avoid typing it twice here and the instance below, I think this could be

```ts
t.expression(() => note = factory.GraceNote(noteStruct));
```

jaredjj3 (approved):
Thanks for adding support for this! Now that we have some depth, do you see a better way to maintain the `producer` and `renderer` modules?

src/renderer.ts
rvilarl: You would need to define notestruct then. I will leave it as it. I think we have to look at the output code anyway, it is getting too complex.

src/types.ts
rvilarl: Done

src/renderer.ts
rvilarl: Done

src/types.ts
rvilarl: OK

PR #30, opened 2022-04-15
Improve dev app

jaredjj3:
This PR:

- adds all the lilypond examples that I manually updated to version 4.0 MusicXML
- creates a screen-responsive alphabetical index on the landing page that makes it easy to navigate through a lot of examples

image


rvilarl (approved):
I have to learn react :)

rvilarl:
@jaredjj3 it looks much better! 
I have also pushed a fix for the width discussed in 
https://github.com/0xfe/vexflow/issues/1363
together with an increase on the visible part of the score 1000 -> 2000

PR #32, opened 2022-04-16
accidental cautionary support

rvilarl:
This fixes test 01f (more accidentals types added) and adds support for cautionary accidentals.


rvilarl:
I will merge the two open PRs they should resolve a few fail tests

PR #34, opened 2022-04-16
Add filters to dev app

jaredjj3:
The motivation for this PR was to address the poor performance when rendering the index page. This was partly due to the number of renders that the `<AlphabeticalIndex>` component had to do every time a render status changed. Now, the `<AlphabeticalIndex>` component is rendered after all of the examples finish rendering.

- adds `success`, `failed`, and `slowest` filters to the dev app
- adds stats to the dev app
- improves performance on the initial load
- saves settings in `LocalStorage`, so filters persist on hot reload

>NOTE: Toggling the "success" examples is still laggy. I think this is due to 100+ elements being updated at once. I'm not planning to fix this since it's a minor inconvenience.

image

### DIFF
--- a/dev/src/components/Examples.tsx
+++ b/dev/src/components/Examples.tsx
@@ -1,21 +1,47 @@
-import { Divider, Typography } from 'antd';
-import React, { useCallback, useState } from 'react';
+import { Checkbox, Divider, Typography } from 'antd';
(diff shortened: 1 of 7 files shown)


rvilarl (approved):
I trust your change :)

PR #37, opened 2022-04-17
multiline support

### LINKED ISSUE #31: Next features: Accidentals, Clefs, scores on several lines
I believe the next features should be to support all the combinations present in the tests for:
- accidentals, including parenthesis
- clefs

Support multiple lines for scores.

@jaredjj3 do you have another preference?

rvilarl:
fixes #31 

With this one we can close #31. Now the scores are complete.

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -63,6 +63,8 @@ export class Renderer {
     let voice = t.let('voice', () => 0);
     let x = t.let('x', () => 0);
     let y = t.let('y', () => 0);
+    let yCur = 100;
+    let xMax = 0;
     let width: number | undefined = t.let('width', () => undefined);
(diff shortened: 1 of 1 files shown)


src/renderer.ts
```diff
@@ -248,6 +251,13 @@ export class Renderer {
           );
           t.expression(() => factory.draw());
           t.expression(() => (x += staves.get(1)!.getWidth()));
+          if (x > 1500) {
```
jaredjj3, line 254: Seems like the renderer needs to know about its container's context. We will need to use [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) and re-render on resize.

jaredjj3 (approved):

PR #40, opened 2022-04-22
notations & beam refactored

rvilarl:
A lot of progress with notations, I have also refactored to reduce the number of messages.

image

### DIFF
--- a/src/producer.ts
+++ b/src/producer.ts
@@ -120,24 +120,53 @@ export class Producer {
           };
           messages.push(this.lastNoteMessage);
         }
+
+        const getNodeDetails = (
+          node: Node
+        ): { name: string; value: string; number: number; type?: string; placement?: string } => {
+          const nodeElem = node as Element;
+          return {
+            name: node.nodeName,
+            value: node.textContent ?? '',
+            number: parseInt(nodeElem.getAttribute('number') ?? '1'),
+            type: nodeElem.getAttribute('type') ?? undefined,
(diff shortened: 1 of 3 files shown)


jaredjj3 (approved):
Thanks for cleaning up these features. I'm glad you were able to simplify things a little bit. Do you think we're at a point where we can start writing unit tests and rethink the overarching approach?

rvilarl:
Yes we should. From an overarching perspective we definitively need some structure otherwise the renderer will be a mess. I thought that the plan was to base the testing on the snapshots and the unofficial tests.

jaredjj3:
>I thought that the plan was to base the testing on the snapshots and the unofficial tests.

I'll defer to your judgment if this is the best thing to do since the core library is mostly your work.

We can continue to do that, but the way we've been working can really only support one developer working on it (or otherwise we face merge conflict hell). You've built up a lot of momentum developing the core library, and now that we have something tangible, `vexml` might be easier to "sell" as a real possible MusicXML -> Vexflow solution and get more developers on board. Personally, I'd like the library to be in a place where more than one developer can work on it, but I don't really have a good sense of how far we are from that.

>I am very happy because I am detecting a lot of issues in VexFlow with this. 🙂

I'm genuinely happy to hear this. This is one of the things that I love about open-source software. I also think this is an avenue where we could get more support on this library.

rvilarl:
I am happy to get more structure to allow more people to cooperate. What would you suggest?

PR #41, opened 2022-04-22
Improve snapshotting

jaredjj3:
- adds snapshots for all the examples
- automatically uploads a snapshot if the example is successfully rendered and a server snapshot is missing
- augments the `Snapshot` class to be able to compare
- resizes images to be comparable if their dimensions don't match (without stretching)
- eagerly loads the snapshot data in the index


jaredjj3:
Just going to merge this while I think about the maintainability problem more.

rvilarl:
I am looking into how to resolve the chords in a sensible way. I want to go for scores with several parts then. I believe that both changes may have a big impact.

PR #44, opened 2022-04-28
simple time signatures support

### LINKED ISSUE #43: Key signatures missing.
@jaredjj3 I will go for this one as chords and parts will take longer and I believe that this is an important inforamation that is missing in our representations.

rvilarl:
Fixes #43 
The standard ones are there. :)

### DIFF
--- a/src/producer.ts
+++ b/src/producer.ts
@@ -51,7 +51,7 @@ export class Producer {
         break;
       case 'attributes':
         {
-          const message: AttributesMessage = { msgType: 'attributes', clefs: [] };
+          const message: AttributesMessage = { msgType: 'attributes', clefs: [], times: [], keys: [] };
           const clefElems = nodeElem.getElementsByTagName('clef');
           for (let i = 0; i < clefElems.length; i++) {
             const staff = parseInt(clefElems.item(i)!.getAttribute('number') ?? '1');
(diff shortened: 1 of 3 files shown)


src/producer.ts
```diff
@@ -66,11 +66,17 @@ export class Producer {
...
           }
-          const fifths = nodeElem.getElementsByTagName('fifth').item(0)?.textContent;
-          if (fifths) message.key = parseInt(fifths);
-          const beats = nodeElem.getElementsByTagName('beats').item(0)?.textContent;
-          const beatType = nodeElem.getElementsByTagName('beat-type').item(0)?.textContent;
-          if (beats && beatType) message.time = `${beats}/${beatType}`;
+          const keyElems = nodeElem.getElementsByTagName('key');
+          for (let i = 0; i < keyElems.length; i++) {
+            const fifths = parseInt(keyElems.item(i)!.getElementsByTagName('fifths').item(0)?.textContent ?? '0');
+            message.keys.push({ fifths });
```
jaredjj3, line 72: You need to also include the staff number in the message.

src/producer.ts
```diff
@@ -66,11 +66,17 @@ export class Producer {
...
+          }
+          const timeElems = nodeElem.getElementsByTagName('time');
+          for (let i = 0; i < timeElems.length; i++) {
+            const beats = timeElems.item(i)!.getElementsByTagName('beats').item(0)?.textContent;
+            const beatType = timeElems.item(i)!.getElementsByTagName('beat-type').item(0)?.textContent;
+            if (beats && beatType) message.times.push({ signature: `${beats}/${beatType}` });
```
jaredjj3, line 78: You need to also include the staff number in the message.

jaredjj3 (changes requested):

src/producer.ts
rvilarl: I was too lazy as it was not needed for basic functionality, :). Done.

PR #47, opened 2022-05-06
basic part support

### LINKED ISSUE #36: Multiple parts
@jaredjj3 I just pushed a minor commit to support breve and long notes. There are only two tests failing and they are related to multiple parts scored which still not supported.

There are two work arounds for VexFlow bugs documented with 'VexFlow bug:' in the code

rvilarl:
Fixes #36

Only basic support. We should start to divide renderer.ts in different files

### DIFF
--- a/src/producer.ts
+++ b/src/producer.ts
@@ -34,7 +34,7 @@ export class Producer {
     while (this.partsIndex[0] < this.parts[0].childs.length) {
       const messages: EasyScoreMessage[] = [];
       for (let i = 0; i < this.parts.length; i++) {
-        messages.push({ msgType: 'partStart', id: this.parts[i].id, ndx: i + 1, from: this.parts.length });
+        messages.push({ msgType: 'partStart', msgIndex: i, msgCount: this.parts.length, id: this.parts[i].id });
         do {
           messages.push(...this.getMessages(this.parts[i].childs.item(this.partsIndex[i])));
           this.partsIndex[i] += 1;
@@ -42,7 +42,7 @@ export class Producer {
           this.parts[i].childs.item(this.partsIndex[i] - 1).nodeName != 'measure' &&
           this.partsIndex[0] < this.parts[0].childs.length
         );
-        messages.push({ msgType: 'partEnd', id: this.parts[i].id, ndx: i + 1, from: this.parts.length });
+        messages.push({ msgType: 'partEnd', msgIndex: i, msgCount: this.parts.length, id: this.parts[i].id });
       }
       for (const message of messages) {
         receiver.onMessage(message);
@@ -172,7 +172,6 @@ export class Producer {
 
         // Notations
         const notations = nodeElem.getElementsByTagName('notations').item(0)?.childNodes;
-        let index = 0;
         for (let i = 0; notations && i < notations.length; i++) {
           const nodeName = notations.item(i).nodeName;
 
@@ -184,18 +183,15 @@ export class Producer {
                 if (notations.item(i).childNodes.item(j).nodeName !== '#text') {
                   messages.push({
                     msgType: 'notation',
-                    index,
                     ...getNodeDetails(notations.item(i).childNodes.item(j)),
                   });
-                  index++;
                 }
               }
               break;
             case '#text':
               break;
             default:
-              messages.push({ msgType: 'notation', index, ...getNodeDetails(notations.item(i)) });
-              index++;
+              messages.push({ msgType: 'notation', ...getNodeDetails(notations.item(i)) });
               break;
           }
         }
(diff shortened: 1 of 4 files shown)


src/producer.ts
```diff
@@ -29,12 +31,36 @@ export class Producer {
         receiver.onMessage(message);
       }
     }
+    while (this.partsIndex[0] < this.parts[0].childs.length) {
+      const messages: EasyScoreMessage[] = [];
+      for (let i = 0; i < this.parts.length; i++) {
+        messages.push({ msgType: 'partStart', id: this.parts[i].id, ndx: i + 1, from: this.parts.length });
```
jaredjj3, line 37: nit: It's not an index if you're adding 1 like this. Would you either rename the property as "number" or remove the `+ 1` please?

src/producer.ts
```diff
@@ -29,12 +31,36 @@ export class Producer {
...
 
   private getMessages(node: Node): EasyScoreMessage[] {
     const messages: EasyScoreMessage[] = [];
     const nodeElem = node as Element;
     switch (node.nodeName) {
+      case 'part':
+        {
+          const id = nodeElem.getAttribute('id') ?? 'NN';
+          this.parts.push({ id, childs: node.childNodes });
+          this.partsIndex.push(0);
```
jaredjj3, line 61: This doesn't seem to be doing anything. Do we need `this.partsIndex` at all?

src/renderer.ts
```diff
@@ -96,17 +97,69 @@ export class Renderer {
...
+            const notesWidth =
+              staveWidth -
+              staves.get(`${curPart}_1`)!.getNoteStartX() +
+              staves.get(`${curPart}_1`)!.getX() -
+              VF.Stave.defaultPadding;
+            formatter!.format(voicesArr, notesWidth);
+            t.literal(`formatter.format(voicesArr, ${notesWidth});`);
+            t.expression(() => factory.draw());
+            x += staves.get(`${curPart}_1`)!.getWidth();
+            //t.expression(() => voices.clear());
```
jaredjj3, line 137: nit: Remove

src/renderer.ts
```diff
@@ -96,17 +97,69 @@ export class Renderer {
...
+            //t.expression(() => staves.clear());
+            if (x > 1500) {
```
jaredjj3, line 139: It seems like we need to formalize what the context's width should be. Later on, I think we need to let the renderer know what width to use. For now, do you mind lifting all the values like this into named constants so we can see the relationship between all of them more clearly? Feel free to do that in a downstream PR.

src/types.ts
```diff
@@ -4,6 +4,13 @@ export type MeasureStartMessage = {
   staves?: number;
 };
 
+export type PartStartMessage = {
+  msgType: 'partStart';
+  id: string;
+  ndx: number;
+  from: number;
```
jaredjj3, line 11: nit: Based on its usage, this `from` property references the `ndx`. Would you indicate that in the property name?

src/types.ts
```diff
@@ -16,6 +23,13 @@ export type MeasureEndMessage = {
   msgType: 'measureEnd';
 };
 
+export type PartEndMessage = {
+  msgType: 'partEnd';
+  id: string;
+  ndx: number;
+  from: number;
```
jaredjj3, line 30: nit: Based on its usage, this `from` property references the `ndx`. Would you indicate that in the property name?

jaredjj3 (changes requested):
IMO, this is a huge improvement. Thanks for adding this.

src/types.ts
rvilarl: Waht name do you recommend? ndx will be number

rvilarl: maxNumber, total, ??

src/producer.ts
rvilarl: partsIndex is used above on lines 39 & 43. They are used to keep track on how far we wentwith the messages on each part for each of the measures.

src/types.ts
rvilarl: I went for msgIndex and msgCount.

src/renderer.ts
rvilarl: At want stage I will add support for print messages. MusicXML dictates the width of the page.

src/producer.ts
jaredjj3: SG

src/renderer.ts
jaredjj3: SG

jaredjj3 (changes requested):
LGTM, thanks for adding this.

jaredjj3 (approved):

PR #48, opened 2022-05-07
new class: notations

rvilarl:
This is a possible way of reducing complexity. What do you think?

### DIFF
--- a/src/notations.ts
+++ b/src/notations.ts
@@ -1,0 +1,201 @@
+import * as VF from 'vexflow';
+import { CodeTracker, NotationMessage } from './types';
+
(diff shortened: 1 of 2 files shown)


jaredjj3:
I see how this could work. The `Renderer` can be composed of other renderers. The challenge is managing shared state. How do you see that working?

rvilarl:
If we pass the shared elements as parameters like notes, it will force us to keep it to a minimum, will not it?

jaredjj3:
Passing state as params makes sense to me. It clearly communicates what the dependencies are. Feel free to evolve this idea.

PR #49, opened 2022-05-12
new class: attributes

rvilarl:
Message attributes handled in a separate class


jaredjj3 (approved):
Looks like a standard lift-and-shift. Thanks for doing this!

Issue #51, opened 2022-05-17
Inhibition of messages


rvilarl:
Hi Jared, I will make later today a PR to start handling the lyrics. The lyrics will hide problems with minimal distance formatting. I thought that it might be good to let the user inhibit elements and/or messages on the GUI. What do you think?

jaredjj3:
>on the GUI

Just to clarify, do you mean in the dev app?

I think that the MusicXML document is the sole source of truth. If a user doesn't want something rendered, they should update the MusicXML document and feed it into `vexml`. Another approach is that they can add a middleware to the renderer and filter the messages themselves, but this seems dangerous since they could filter/inject/mutate messages in a way that the renderer (or any other message receiver) is not expecting.

rvilarl:
Yes I meant in the dev app. I will comment on the inhibition on the PR.

jaredjj3:
This is obsolete since we moved off the state machine approach.

PR #52, opened 2022-05-17
LyricMessage support

rvilarl:
Basic support for lyrics. Yo will see that in 32a-Notations there are no overlaps

32a-Notations
-

### DIFF
--- a/src/producer.ts
+++ b/src/producer.ts
@@ -110,6 +110,11 @@ export class Producer {
           messages.push(message);
         }
         break;
+      case 'lyric':
+        const text = nodeElem.getElementsByTagName('text').item(0)?.textContent ?? '';
+        const syllabic = nodeElem.getElementsByTagName('syllabic').item(0)?.textContent ?? 'single';
+        messages.push({ msgType: 'lyric', text, syllabic });
+        break;
       case 'note':
         const rest = nodeElem.getElementsByTagName('rest').length > 0;
         const grace = nodeElem.getElementsByTagName('grace').length > 0;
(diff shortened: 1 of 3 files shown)


rvilarl:
About #51 I believe that someone might want to print a score without lyrics (instrumental only) and without having to edit the musicxml file.

jaredjj3:
Awesome work! That looks good. I'll review shortly.

As far as configuration goes, I think we should only allow configuring things that MusicXML can't accomplish. A parallel is `opensheetmusicdisplay's` [EngravingRules](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/69ff4a62f4a2c487cfb91aed46eec5b617b5749e/src/MusicalScore/Graphical/EngravingRules.ts) class, but I believe this actually clashes with the MusicXML spec. I had a hard time using this and deciding whether or not to make a change in MusicXML or in the engraving rules.

I think we should push that complexity on the user, since the concern of excluding some notations belongs to the MusicXML document in my opinion. Knowing that the MusicXML spec is very complex, we can allow users to imperatively filter messages by adding an option to `RendererOptions`:

```ts
type RendererOptions = {
  codeTracker?: CodeTracker;
  UNSAFE_shouldProcessMessage?: (message: EasyScoreMessage) => boolean;
};
```

I think that's along the lines of what you originally proposed.

jaredjj3 (approved):
This PR being simpler is a good sign we're headed in the right direction. :)

jaredjj3:
Actually, I like this better:

```ts
type RendererOptions = {
  codeTracker?: CodeTracker;
  UNSAFE_transformMessage?: (message: EasyScoreMessage, ctx: any) => EasyScoreMessage[] | EasyScoreMessage | null;
};
```

PR #53, opened 2022-05-19
BarlineMessage support

rvilarl:
There is some work to do in VexFlow to get all the possibilites right and this can also be separated on an independent class.

However I believe that it can be merged as a starting point. :)


jaredjj3 (approved):
LGTM, thank you for taking the time to add this!

PR #55, opened 2022-06-03
more noteheads supported

rvilarl:
Some progress :)

Changes will be visible once https://github.com/0xfe/vexflow/pull/1401 is merged in vexflow.


jaredjj3 (approved):
Awesome! Do you know if we have coverage in the dev app examples for these new noteheads?

rvilarl:
It is basically test 22a. I believe that coverage is not complete in both directions. There are still noteheads missing in my PR in Vexflow. The good news is that we will reach a higher compliance than the one shown in the Unofficial tests.🙂

PR #65, opened 2022-09-29
render Clefs on new lines

rvilarl:
Octave changes are not being kept this requires a change in VexFlow as referenced in the issue.
I have started writing comments :)


rvilarl:
The last line is not representing the treble staff one octave above.
image

rvilarl:
I forced an update to avoid the changes caused by eslint. This should facilitate the review.

jaredjj3 (approved):
Looks good to me, thanks for adding this.

PR #68, opened 2022-10-02
Refactor of producer module (preview)

jaredjj3:
Hey @rvilarl, I've been cleaning up the producer modules and I developed this new pattern that prioritizes testability and encapsulation. I haven't wired it into the main flow of things, so anything you're working on should be unaffected for now.

We can now reap the benefits that `VexmlMessage` (formerly `EasyScoreMessage`) gives us:

- We can develop producer and rendering features separately.
- It's easier to test the producers using POJOs instead of evaluating image diffs or evaluating the equality of complex objects.

Let me know what you think. If you like it, I'll set up Github actions so we can move towards publishing this library.


src/attributes.ts
```diff
@@ -14,7 +14,7 @@ export class Attributes {
     cur1stStave: number,
     duration: number,
     notes: Array<VF.Note>,
-    systems: VF.System[]
+    systems: System[]
```
rvilarl, line 17: Why is System no longer VF.System and Note stays as VF.Note?

src/notations.ts
```diff
@@ -99,7 +99,7 @@ export class Notations {
           }
           if (modifier.class == 'O') {
             notes[notes.length - 1].addModifier(
-              factory.Ornament(modifier.type, { position: message.placement ?? 'above' }),
+              (factory as any).Ornament(modifier.type, { position: message.placement ?? 'above' }),
```
rvilarl, line 102: Why is 'as any' now required?

src/types.ts
```diff
@@ -107,3 +117,14 @@ export interface CodeTracker {
...
+// TODO(jared): Remove hack by contributing to the types upstream.
+export type System = VF.System & {
+  getStaves(): VF.Stave[];
+  getX(): number;
+  setX(x: number): void;
+  getY(): number;
+  setY(y: number): void;
+  addVoices(voices: VF.Voice[]): void;
+  appendSystem(width: number): System;
+};
```
rvilarl, line 130: Why is that necessary?

rvilarl (commented):
Hi Jared,
I trust that the change is good, I do not get it 100% but go for it! I comment on a few things as well.
Cheers, Rodrigo

src/types.ts
jaredjj3: `VF.System` needs to be updated to include the additional types. Without this, this causing type errors.

src/notations.ts
jaredjj3: To workaround type errors.

src/attributes.ts
jaredjj3: VF.System needs to be updated to include the additional types. Without this, there will be type errors.

src/types.ts
rvilarl: All you need to do is to install the vexflow pre release under ./vexflow/

rvilarl: `yarn add ./vexflow/vexflow-5fa72362.tgz`

rvilarl: This might also resolve the other two comments

jaredjj3:
> Hi Jared, I trust that the change is good, I do not get it 100% but go for it! I comment on a few things as well. Cheers, Rodrigo

In the current `Producer`, you need to coordinate state inside the class and extend `Producer.getMessages` to handle new nodes. This is hard to maintain, especially at the rate we're going. I made the following changes:

- Move general state into a `ctx` variable that can be passed around. I think this will be easier to reason about and test, since you don't have to prime the class with certain state.
- Create the concept of node handlers that translate `Node` objects into `VexmlMessages`. This makes the inputs/outputs clear: given a node, you expect certain messages to be produced. This makes it easy to test. At first, I wanted to do simple functions, but it got messy trying to fit all the logic in a simple function.
- Use a lightweight dependency injection system by letting the child handlers be passed into the constructor. This makes it possible to truly isolate testing a given handler — instead of injecting a real handler, you inject a `NoopHandler` and then use the testing framework to spy/mock out the methods of interest.

The theme I was going for was encapsulation and testability. Hopefully, it'll make more sense as I add more handlers.

PR #73, opened 2022-11-26
Refactor Vexml message producers

jaredjj3:
This PR replaces the `legacyproducer` module with the `producer` module.

Vexml message producers are fragmented to make them easier to develop and test. They are not strictly separated by XML elements, but instead by semi-logical chunks. These chunks will be further refined in a downstream PR.

The refactor is not 100% done, but it's at a point where I'd like to merge it in to avoid conflicts.


rvilarl:
Wow a big change, I trust you, merge it :)

PR #75, opened 2022-12-10
Organize the src dir

jaredjj3:
Having a flat dir structure won't work out too well since there are still a lot of cases we need to cover. I wanted to push this out before I refactored more, to avoid merge conflicts on your end.


rvilarl (approved):

rvilarl:
@jaredjj3 lgtm Thanks!

PR #77, opened 2023-01-15
Vexflow beta2 & fermata fix

rvilarl:
Some progress.


jaredjj3 (approved):
Looks good! I think I will take some time to improve the typing on the producer side, so you can have better type hints when developing things like this.

If any of the diffs changed, would you take a snapshot and include it in this PR please?

rvilarl:
**previous**
32a-Notations
**with this PR**
32a-Notations

PR #79, opened 2023-01-19
joplin score added

rvilarl:
Joplin score added and passacaglia renamed


jaredjj3:
I'm glad we have this feature. Thanks for improving it with scores that encompass a lot of musicxml features!

PR #80, opened 2023-01-19
beam auto stem fix

rvilarl:
first fix, score looking much better, I believe. :)

AAb_Joplin_Gladioulus_Rag


jaredjj3 (approved):
LGTM, thanks for fixing this!

PR #81, opened 2023-01-20
fix notes with dots duration

rvilarl:
Measure 5 now aligned properly. StaveNotes were not considering the additional duration cause by the dots.

AAb_Joplin_Gladioulus_Rag


jaredjj3 (approved):
LGTM

PR #84, opened 2023-02-10
printHandler added

rvilarl:
@jaredjj3 Joplin is starting to look good!


rvilarl:
@jaredjj3 I have added a test but I do not feel confident with tests. Could you complete it?

This below is an example. The spec is in https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/print/
```xml
<print new-system="yes">
<system-layout>
<system-margins>
<left-margin>21.00</left-margin>
<right-margin>-0.00</right-margin>
</system-margins>
<system-distance>165.00</system-distance>
</system-layout>
<staff-layout number="2">
<staff-distance>80.00</staff-distance>
</staff-layout>
</print>
```

jaredjj3 (approved):
LGTM

jaredjj3:
> @jaredjj3 I have added a test but I do not feel confident with tests. Could you complete it?
> 
> This below is an example. The spec is in https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/print/
> 
> ```
> <print new-system="yes">
> <system-layout>
> <system-margins>
> <left-margin>21.00</left-margin>
> <right-margin>-0.00</right-margin>
> </system-margins>
> <system-distance>165.00</system-distance>
> </system-layout>
> <staff-layout number="2">
> <staff-distance>80.00</staff-distance>
> </staff-layout>
> </print>
> ```

Sure, I'm happy to add tests! Thanks for the PR. Joplin is looking better!

rvilarl:
Thanks! I will Look next into the curves and after I will check the generated code. 😃

PR #85, opened 2023-02-12
code generation fixes (for Joplin)

rvilarl:
Generated code and curves are looking better

https://jsfiddle.net/v53kesbp/2/


src/receivers/notations.ts
```diff
@@ -13,14 +13,22 @@ export class Notations {
...
           if (Notations.slurStart[message.number - 1] >= 0) {
+            const bSameDir =
+              notes[Notations.slurStart[message.number - 1]]?.getStemDirection() ==
+              notes[notes.length - 1]?.getStemDirection();
             factory.Curve({
               from: notes[Notations.slurStart[message.number - 1]],
               to: notes[notes.length - 1],
-              options: {},
+              options: {
+                invert: bSameDir ? true : false,
```
jaredjj3, line 23: nit: Just set this to `bSameDir`, which looks like it's already a boolean.

jaredjj3 (approved):
LGTM

PR #86, opened 2023-02-13
Retire attributes message

jaredjj3:
@rvilarl, now that you've gained a lot of momentum on the receiver side, I'm going to start studying it and figure out a maintainable long-term solution. First, I need to clean up the messages.

This PR is a non-invasive change that "flattens" the `AttributesMessage`, and typechecks some string values. I'm also starting to replace most (if not all) usages of `undefined` in messages with `null`, because `undefined` fields are omitted when serializing to JSON.


rvilarl:
LGTM

PR #90, opened 2023-06-07
Refactor the project

jaredjj3:
Initially, we thought that `vexml` could be a state machine for MusicXML. However, this was fundamentally incompatible with `vexflow`. **We were constantly creating brittle bridges between how music data is stored vs how it is rendered.**

The new approach ditched the state machine/event-based approach for a more imperative one. This unlocks the flexibility we need to interface with `vexflow`.

This PR refactors `vexml` completely, using the strategies from the old approach:

- I created first-class objects that describe different elements in MusicXML. This is different than the approach in [musicxml](https://github.com/stringsync/musicxml). `musicxml` was just a TypeScript wrapper around the XML elements, while the objects created in this PR have useful abstractions. For example, the `<note>` element has the `Note` class, which has `Note.getChordTail`, which returns all subsequent the notes after it that compose the chord.
- I unit-tested almost everything. The only thing that's not well tested is the `Vexml` class, which we're currently relying on the dev app at the moment.
- I improved rendering speed by about 66% (results may vary).
- I added support for automatically formatting systems into separate lines. Systems will also be stretched so that there is no excess length per line.


jaredjj3:
@rvilarl, I wasn't able to get to full parity, because I kept running into roadblocks using `vexflow`. However, almost all of the examples render. Would you review and let me know if you'd like to merge this? If not, that's completely ok — just let me know what you'd like to see done before you're ok with merging.

We should also consider forking `master` so that we can continue to _easily_ use it as a reference when restoring the missing features.

rvilarl:
Thanks a lot 66% quicker ist awesome. I will review as soon as I can!

rvilarl (approved):
Interesting to see that the unofficial musicxml test set has bugs (double-flat).
Thanks a lot, looks very good!

rvilarl:
@jaredjj3 there are conflicts that need to be resolved before merging.

jaredjj3:
Thanks for reviewing, I resolved the merge conflicts, and I'm going to move forward with the merge. The legacy codebase is in this branch [vexml/legacy](https://github.com/stringsync/vexml/tree/legacy), so feel free to bookmark it for convenience.

We'll need to build back up some features, but I think this will scale much better than our previous approach.

PR #91, opened 2023-08-24
Fix beaming

jaredjj3:
I removed beam post format before anything was drawn.

Before:

image

After:

image


jaredjj3:
@rvilarl I think this is a small enough request to just merge. I see you're busy in vexflow/vexflow, so let me know if/when you have capacity for this again.

rvilarl:
Thanks a lot

PR #93, opened 2023-08-25
Update README with instructions on how to run

jaredjj3:


rvilarl:
Thanks Jared, good to see you back. VexFlow5 is progressing. I will integrate it once it is in a decent status.

PR #96, opened 2023-09-12
Establish rendering hierarchy

jaredjj3:
This PR splits the rendering logic from `vexml.ts` into a hierarchy that is a coalescence of the following:

- MusicXML
- VexFlow
- Music Theory

It took a while to realize that the three "API" surfaces do not align, which seems to warrant the need for an additional abstraction.

This PR also creates a mechanism to give callers a programmatic handle on what's been rendered via the `*Rendering` types.


jaredjj3:
@rvilarl this PR is a breakthrough and I feel that there is a low chance I'll need to do a major refactor after merging. The problem was that I saw [vexml.ts](https://github.com/stringsync/vexml/blob/a51985c118dd26365eb82b7c8817fed81dea33b4/src/vexml.ts) will grow in an uncontrollable manner. I tried taming it by wrapping some `vexflow` objects in my own, but it felt like a bandaid solution.

This PR breaks up the rendering pipeline using familiar musical hierarchy, and I'm finally satisfied with the maintainability trajectory with this library.

After this PR is merged, the next goal is to flesh out the Github project with the lilypond test suite. I think we'll need other test suites, but I figured this might be a good start.

I expect this PR to be merged by the end of this week. Please let me know if you see anything concerning with the progress so far.

rvilarl:
@jaredjj3 thanks a lot, no concerns, it is looking good. I will in the next few weeks update to VexFlow5.

PR #98, opened 2023-09-17
vexflow5

rvilarl:
@jaredjj3 this is the one

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "test:ci": "docker run --rm vexml:latest yarn jest --runInBand --ci"
   },
   "dependencies": {
-    "vexflow": "4.2.3"
+    "vxflw-early-access": "5.0.0-alpha.2"
   },
   "devDependencies": {
     "@babel/core": "^7.17.8",
--- a/src/rendering/accidental.ts
+++ b/src/rendering/accidental.ts
@@ -1,5 +1,5 @@
 import * as musicxml from '@/musicxml';
-import * as vexflow from 'vexflow';
+import * as vexflow from 'vxflw-early-access';
 
 /** The result of rendering an Accidental. */
 export type AccidentalRendering = {
--- a/src/rendering/beam.ts
+++ b/src/rendering/beam.ts
@@ -1,5 +1,5 @@
 import * as musicxml from '@/musicxml';
-import * as vexflow from 'vexflow';
+import * as vexflow from 'vxflw-early-access';
 
 /** The result of rendering a Beam. */
 export type BeamRendering = {
--- a/src/rendering/note.ts
+++ b/src/rendering/note.ts
@@ -1,5 +1,5 @@
 import * as musicxml from '@/musicxml';
-import * as vexflow from 'vexflow';
+import * as vexflow from 'vxflw-early-access';
 import { Beam } from './beam';
 import { Accidental, AccidentalRendering } from './accidental';
 import { Config } from './config';
--- a/src/rendering/rest.ts
+++ b/src/rendering/rest.ts
@@ -1,5 +1,5 @@
 import * as musicxml from '@/musicxml';
-import * as vexflow from 'vexflow';
+import * as vexflow from 'vxflw-early-access';
 import { Config } from './config';
 
 /** The result of rendering a Rest. */
--- a/src/rendering/score.ts
+++ b/src/rendering/score.ts
@@ -1,6 +1,6 @@
 import { System, SystemRendering } from './system';
 import * as musicxml from '@/musicxml';
-import * as vexflow from 'vexflow';
+import * as vexflow from 'vxflw-early-access';
 import * as util from '@/util';
 import { Config, DEFAULT_CONFIG } from './config';
 
--- a/src/rendering/stave.ts
+++ b/src/rendering/stave.ts
@@ -1,5 +1,5 @@
 import * as musicxml from '@/musicxml';
-import * as vexflow from 'vexflow';
+import * as vexflow from 'vxflw-early-access';
 import { Voice, VoiceRendering } from './voice';
 import { Config } from './config';
 import * as util from '@/util';
--- a/src/rendering/voice.ts
+++ b/src/rendering/voice.ts
@@ -1,5 +1,5 @@
 import * as musicxml from '@/musicxml';
-import * as vexflow from 'vexflow';
+import * as vexflow from 'vxflw-early-access';
 import { Note, NoteRendering } from './note';
 import { Chord, ChordRendering } from './chord';
 import { Rest, RestRendering } from './rest';
--- a/yarn.lock
+++ b/yarn.lock
@@ -5821,10 +5821,10 @@ v8-to-istanbul@^9.0.1:
     "@types/istanbul-lib-coverage" "^2.0.1"
     convert-source-map "^1.6.0"
 
-vexflow@4.2.3:
-  version "4.2.3"
-  resolved "https://registry.yarnpkg.com/vexflow/-/vexflow-4.2.3.tgz#2db536297fc97fd1f502cdc042ccc6caefd9bc11"
-  integrity sha512-YiwYhZYuXPKqPfvvIC0QRYS3xnInDjeAxJeAZsQ+njTEE3q10sZMfo7RRGXj40rg7n74AFWdTiN3FymW8qE1Jw==
+vxflw-early-access@5.0.0-alpha.2:
+  version "5.0.0-alpha.2"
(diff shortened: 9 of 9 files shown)


rvilarl:
I am lost now with docker... of course it fails

jaredjj3:
In tsconfig, I think you can map imports from `vexflow` to `vxflw-early-access`.

```json
{
  "compilerOptions": {
    "paths": {
      "vexflow": ["node_modules/vxflw-early-access"]
    }
  }
}
```

So, revert the `import` statements and try this mapping... Annoyingly you might need to update jest.config.js too:

```js
module.exports = {
  moduleNameMapper: {
    '@/(.*)': '<rootDir>/src/$1',
    'vexflow': '<rootDir>/node_modules/vxflw-early-access',
  },
};
```

I'm unsure if that's needed or if that'll work.

jaredjj3:
> I am lost now with docker... of course it fails

Does it work on `main` at `HEAD`? If not, I'll consider making a testing server for us to have a consistent environment to test against. The complication is that Google Chrome (which is what `puppeteer` is using for the image snapshot tests) renders SVG slightly differently depending on the platform.

You can always just run the tests locally with `yarn jest`, but expect the integration tests to fail (unless you're on the same platform as the Docker image).

rvilarl:
can you try to get this one to run? yarn test on master runs ok

jaredjj3:
Sure, I'm just finishing some docs, then I can look at this in an hour or so.

jaredjj3:
I was able to get the tests to run in https://github.com/stringsync/vexml/commit/f2861b1e70b5181a4adb853749f4b47c1cd0f24c. However, the integration tests fail with:

```
[RuntimeError] Font: No txt context
```

Would you copy that commit and see if you get the same thing?

rvilarl:
@jaredjj3 how do I pull this commit from github? I have no idea.

rvilarl:
I also do not remember which yarn command was the right one to see the test results in the browser. Too much time with VexFlow

rvilarl:
> Font: No txt context

@jaredjj3 this relates to not being able to create a canvas to measure text. Does the react infraestructre not support canvas? 

```js
 /** Measure the text using the textFont. */
  measureText(): TextMetrics {
    let txtCanvas = Element.#txtCanvas;
    if (!txtCanvas) {
      // Create the canvas element that will be used to measure text in the event
      // of a cache miss.
      txtCanvas = document.createElement('canvas');
      Element.#txtCanvas = txtCanvas;
    }
    const context = txtCanvas.getContext('2d');
    if (!context) throw new RuntimeError('Font', 'No txt context');
    context.font = Font.toCSSString(Font.validate(this.#fontInfo));
    this.#textMetrics = context.measureText(this.#text);
    this.#height = this.#textMetrics.actualBoundingBoxAscent + this.#textMetrics.actualBoundingBoxDescent;
    this.#width = this.#textMetrics.width;
    this.#metricsValid = true;
    return this.#textMetrics;
  }
```

jaredjj3:
>I also do not remember which yarn command was the right one to see the test results in the browser. Too much time with VexFlow

It used to be `yarn dev`. However, when I added the snapshot testing infrastructure, I removed the dev server because I didn't want us to maintain multiple snapshot sources. Someone might come along and wonder which snapshot is correct.

You will need to use `yarn jest integration` (local) or `yarn test integration` (docker) to render the examples wired up in the tests > integration directory. The snapshots get saved to tests > integration > __image_snapshots__ and the diffs (from any test run) are in the __image_snapshots__ > __diff_output__ directory.

If you want to update snapshots, you'll need to run `yarn test integration --updateSnapshot` and you must run it in Docker because it will replicate the environment that Github actions runs with.

I think the testing infrastructure is less convenient, but it standardizes structure for us to achieve `vexml@1.0.0`. As you try to use it, would you let me know what parts don't work or are unclear in the root README? I'll try to update the dev workflow and README accordingly.

jaredjj3:
>@jaredjj3 this relates to not being able to create a canvas to measure text. Does the react infraestructre not support canvas?

React is out of the picture now, and we're using puppeteer to render the examples in a Chrome browser. See `lilypond.test.ts `for how this is done.

I'm not able to do this right now, but the next step to troubleshooting this is to see if our puppeteer setup has issues with canvas. In `lilypond.test.ts`, temporarily comment out the tests and make a new one that creates a `<canvas>` and renders something with it.

(inside a test)

```ts
const canvas = document.createElement('canvas');
canvas.setAttribute('id', 'canvas');

await page.setContent(canvas);

await page.evaluate(() => {
  const canvas = document.getElementById('canvas');
  const ctx = canvas.getContext('2d');
  ctx.fillStyle = 'green';
  ctx.fillRect(50, 50, 100, 100);
});

await page.screenshot({ path: 'foo.png' });
```

If all that works, then puppeteer is not the problem and we need to dig further why this might not be attaching to the DOM.

jaredjj3:
Quick addendum: an even simpler approach is to do everything inside the page.evaluate callback.

jaredjj3:
@rvilarl I was able to take a look and I found the issue. The problem is with jest and jsdom because that's where the `vexflow` APIs are being called.

I reran the tests and got errors like:

```
Error: Not implemented: HTMLCanvasElement.prototype.getContext (without installing the canvas npm package)
```

I followed the instructions on [Automattic/node-canvas](https://github.com/Automattic/node-canvas) and installed canvas locally. I didn't have to do anything with Docker to get it to work.

After I did that, i reran the tests and observed that they render, although the renderings seem broken. Maybe this is something we're doing wrong in this library?

snapshot, diff, test result

lilypond-test-ts-lilypond-01-a-pitches-pitches-xml-900-px-1-snap-diff

rvilarl:
@jaredjj3 thanks! this looking good!
The only thing that is missing now are the fonts, the library uses them directly. This why we are getting the squares. Do you know how to use FontFace with puppeteer / jsdom? The fonts are available in https://cdn.jsdelivr.net/npm/vexflow-fonts@1.0.6/

jaredjj3:
@rvilarl we're closer in https://github.com/stringsync/vexml/commit/64042c43e97cadd19b5cbd7c177a00ea9c0767bb, but something is still wrong. The sizing and formatting now match the snapshot, but I'm still getting the boxes. @ronyeh, I didn't see the `otf` files you referenced in https://github.com/vexflow/vexflow/pull/98#issuecomment-1722452632 — they had different names. What else am I doing wrong?

snapshot, diff, test result

lilypond-test-ts-lilypond-01-a-pitches-pitches-xml-900-px-1-snap-diff

rvilarl:
can you share your branch? I would like to get a look into it.

jaredjj3:
Sure the branch has the same name in this repo: https://github.com/stringsync/vexml

If you develop in this repo or give me access to your fork, I can PR into the branch you've been working on.

jaredjj3:
I think I have an idea what's going on: the browser doesn't have the fonts loaded, which is why we're still getting the squares. However, when we render in jsdom, the fonts **are** there, which is why the spacing seems correct.

jaredjj3:
> I think I have an idea what's going on: the browser doesn't have the fonts loaded, which is why we're still getting the squares. However, when we render in jsdom, the fonts **are** there, which is why the spacing seems correct.

@rvilarl, this was the issue. When also I load the fonts _inside of the browser that puppeteer controls_, it works:

snapshot, diff, test result

image

rvilarl:
I will close this one and merge your branch. Thanks @jaredjj3 !!!

> @rvilarl, this was the issue. When also I load the fonts inside of the browser that puppeteer controls, it works:

@ronyeh it seems that we will be able to text with puppeteer!

ronyeh:
I'm glad you were able to figure it out!

Yes, since puppeteer (and similar testing tools like playwright) use real browsers, there is always a way to load a font and measure text using the standard HTML/CSS/JS APIs.

The hurdle we need to solve is to 1) teach developers how to do it and/or 2) have a super easy (or automated) way for VexFlow to do this upon launching. I imagine most developers will run into the empty boxes issue at some point. I did as well. I just want to make it easy so that end users will never see empty boxes on a score.

The problem with jsdom (vexflow legacy code uses it) is that it doesn't conform exactly to modern web APIs. It uses node canvas internally, and that library requires a different solution for loading fonts ahead of time: registerFont(). It also doesn't support woff2, like all modern browsers do.

jaredjj3:
@ronyeh, if there's a way you can detect whether the fonts you need are loaded ([Document.fonts](https://developer.mozilla.org/en-US/docs/Web/API/Document/fonts)?), you can throw an error that has a brief description and a link to a vexflow wiki doc that has a detailed description+an enumeration of common remedies.

However, this approach would undoubtedly detract from developer experience which seems important to vexflow. I think your automated approach would be better in that regard, but I would recommend at least leaving some sort of warning artifact that the fonts were autoloaded, and that may have performance implications.

PR #100, opened 2023-09-17
Render lilypond01a

jaredjj3:
This PR renders 01a-Pitches-Pitches.xml.

## Divergences

- The default time signature is 4/4 (vexml) instead of cut time (lilypond)
- Measure labels are rendered for every measure and part by default (vexml).
- Measure spacing has 100px buffer above the minimum amount of width needed (vexml).
- Systems have a beginning barline (vexml).
- The time signature is rendered for every system (vexml).

## vexml

lilypond-test-ts-lilypond-01-a-pitches-pitches-xml-900-px-1-snap

## lilypond

image


jaredjj3:
I just noticed an issue where the measure numbers jump from 24 to 31 in the vexml rendering. I'm going to fix that before merging.

jaredjj3:
The issue was that 01a-Pitches-Pitches.xml had the wrong measure numbering and vexml was correctly reflecting that. I fixed this by updating 01a-Pitches-Pitches.xml with the correct numbering.

lilypond-test-ts-lilypond-01-a-pitches-pitches-xml-900-px-1-snap

PR #110, opened 2023-09-28
Render lilypond02a

jaredjj3:
This PR renders 02a-Rests-Durations.xml and fixes dot rendering.

## Divergences

- `vexml` and `gp7` do not clobber the measures. This is better for writability at the expense of readability.
- `osmd` clobbers the measures completely, but I imagine this is configurable.
- I'm unfamiliar with the multi measure rest style in `lilypond`. I'm guessing 3 whole measure rests are more widely recognizable and readable.
- In `gp7`, the dots on the smallest divisions are vertically centered.

## vexml

lilypond-test-ts-lilypond-02-a-rests-durations-xml-900-px-1-snap

## lilypond

image

## osmd

image

## gp7

image


sschmidTU:
Yeah, in OSMD this is a setting that probably shouldn't be true by default, it can be disabled like this:
`osmd.EngravingRules.AutoGenerateMultipleRestMeasuresFromRestMeasures = false;`
image
(Though of course this is a rather academic score, OSMD focuses on practical examples)

(Rendering is still not great here, not sure why it only does a 2-measure rest)

jaredjj3:
Ah thanks for the insight. Glad to see that the option is indeed there.

I imagine that the extra whole measure is a one-off error somewhere. I saw the same behavior in http://lilypond.org/doc/v2.25/input/regression/musicxml/collated-files in 02c-Rests-MultiMeasureRests.xml. I could add an issue to `opensheetmusicdisplay` if you like.

Also I'm curious as to why we're getting different results of dots on 128th notes. Maybe due to different versions of `vexflow`?

jaredjj3:
@sschmidTU quick update, I check lilypond's musicxml and it was incorrect.

Here's osmd with the incorrect XML but correct rendering:

image

Here's osmd with the correct XML and correct rendering:

image

No issue in osmd :)

jaredjj3:
I found why a multi measure rest was being unexpectedly rendered in `osmd`... In #112, I ended up with the same result. I think lilypond's musicxml is [wrong here](https://github.com/stringsync/vexml/blob/c3edbb2e6decba046100ea4cbdce19d05b5be4e8/tests/integration/__data__/lilypond/02a-Rests-Durations.xml#L35-L37). `osmd` was ultimately rendering the correct thing.

sschmidTU:
Thanks, good to know!
Yes, OSMD still uses 1.2.93. It's an unfortunate situation, we'd love to upgrade, but for one the new Vexflow 3+ formatter has some issues, and also we did a lot of customization on our pseudo-fork of 1.2.93 (in form of patches, which would require some work to migrate). See [here](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/discussions/1453)

PR #117, opened 2023-10-01
Render score titles

### LINKED ISSUE #58: Missing: Score title

jaredjj3:
Fixes #58

This PR adds support for rendering a musicxml `<movement-title>` elements when it is available. This will break when the size of the of title exceeds the size of the container.

The complexity comes from us not having direct access to the rendering backend. Normally, I'd just create a `<text>` SVG element and set the [text-anchor](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/text-anchor) property to `middle`. However, this is by design. The same approach won't work for a canvas backend, so it's good that we're manually calculating the positioning.


jaredjj3:
@rvilarl, this PR will fix #58. However, there is much more work that needs to be done. As you come across new title-related issues, please open new ones with details on what `vexml` renders vs what you want.

PR #120, opened 2023-10-02
Render lilypond03b

jaredjj3:
This PR renders 03b-Rhythm-Backup.xml and removes the oversimplification of `musicxml.Measure.getNotes`. Prior to this PR, we were ignoring `<backup>` and `<forward>` elements, which are crucial for specifying polyphonic music with multiple voices.

This PR also adds the concept of children to `NamedElement` and establishes an example of how to fetch multiple different tag names in a typesafe manner in an HTML element collection. This is needed because `<note>`, `<backup>` and `<forward>` are all direct children of `<measure>`.

## Divergences

- `gp7` seems to assume that the `<backup>` should go all the way to the beginning, which is incorrect. I'm not sure what the light gray coloring means.

## vexml

image

## lilypond

image

## osmd

image

## gp7

image


jaredjj3:
@rvilarl, GuitarPro7 seems to handle musicxml poorly. Would you recommend other musicxml rendering software I could use for comparisons? I don't mind spending money.

rvilarl:
Free I would recommend MuseScore

jaredjj3:
Note to self: This was the most difficult unit test to implement for and it added some cruft to how voices are calculated. At the moment, I'm not sure how to break `Voice` down into smaller, more maintainable pieces. For now, I think it's wiser to wait to see how the module evolves before trying to refactor `Voice`.

jaredjj3:
Ah, I think I got it. The abstraction question is not _necessarily_:

>How can I break down voices?

It could be:

>How do I represent groups of voices within the same stave+measure?

Right now, the Voice abstraction doesn't answer this well, which is why it broke the rendering object pattern.

Instead, I'm going to create a new abstraction that represents a group of voices. Maybe I could call it VoiceCluster or Chorus. I'm leaning towards the latter since "cluster" is ambiguous on how it's clustered.

I think this is worth the effort now given the development speed so far. I don't want tech debt to accumulate so early on.

Issue #123, opened 2023-10-03
Unable to run tests


infojunkie:
Running `yarn && yarn test`, I get:
```
[..]
warning " > babel-plugin-styled-components@2.0.7" has unmet peer dependency "styled-components@>= 2".
warning " > eslint-config-standard-with-typescript@21.0.1" has incorrect peer dependency "@typescript-eslint/eslint-plugin@^4.0.1".
warning " > eslint-config-standard-with-typescript@21.0.1" has incorrect peer dependency "eslint@^7.12.1".
warning " > eslint-config-standard-with-typescript@21.0.1" has incorrect peer dependency "typescript@^3.9 || ^4.0.0".
warning "eslint-config-standard-with-typescript > @typescript-eslint/parser@4.33.0" has incorrect peer dependency "eslint@^5.0.0 || ^6.0.0 || ^7.0.0".
warning "eslint-config-standard-with-typescript > eslint-config-standard@16.0.3" has incorrect peer dependency "eslint@^7.12.1".
warning " > eslint-plugin-promise@5.2.0" has incorrect peer dependency "eslint@^7.0.0".
error Could not write file "/vexml/yarn-error.log": "EACCES: permission denied, open '/vexml/yarn-error.log'"
error An unexpected error occurred: "EACCES: permission denied, mkdir '/vexml/node_modules'".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
The command '/bin/sh -c yarn install' returned a non-zero code: 1
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
kratib@kindzadza:~/src/music/vexml$ ls -la
total 408
drwxrwxrwx   8 kratib kratib   4096 Oct  3 10:21 .
drwxrwxr-x 161 kratib kratib   4096 Sep 30 16:08 ..
[..]
drwxrwxr-x 509 kratib kratib  20480 Oct  3 10:21 node_modules
```

jaredjj3:
It looks like there are two issues:

1. It looks like there's permission problem creating a `node_modules` directory. Did you try running `sudo yarn`?
2. You also have unmet peer dependencies, but I imagine this might be due to the first issue, since you're unable to install any packages.

infojunkie:
[...]

PR #130, opened 2023-10-06
Fix docker-based test running #123

infojunkie:
- Fix permissions of `/vexml` folder tree to allow installing packages and writing files
- Fix `package.json` `test` script for volume-mounting syntax that is compatible with Linux


jaredjj3:
Thanks again!

PR #131, opened 2023-10-07
Migrate staff* references to stave*

jaredjj3:
This PR updates terms that use "staff" to use "stave" instead. Direct-ish references to musicXML were left untouched.

https://en.wikipedia.org/wiki/Staff_(music)

When I initially studied musicXML and vexflow, I didn't realize these terms were interchangeable. I thought "stave" / "staves" were the plural of "staff", so I was confused when I saw "stave" and "staff" used interchangeably. Hopefully this saves someone else the headache.

Good thing I used TypeScript.

### DIFF
--- a/src/musicxml/attributes.ts
+++ b/src/musicxml/attributes.ts
@@ -2,7 +2,7 @@ import { Clef } from './clef';
 import { Key } from './key';
 import { NamedElement } from '@/util';
 import { Time } from './time';
-import { StaffDetails } from './staffdetails';
+import { StaveDetails } from './stavedetails';
 import { MeasureStyle } from './measurestyle';
 
 /**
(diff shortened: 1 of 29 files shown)


sschmidTU:
Just curious, judging from the Wikipedia article, why use "stave" then when "staff" is more common both in UK and US, and "stave" is UK only? I guess because that makes the plural easier/more similar ("stave" + "staves")?

> the staff (US and UK)[[1]](https://en.wikipedia.org/wiki/Staff_(music)#cite_note-collins-1)[[2]](https://en.wikipedia.org/wiki/Staff_(music)#cite_note-mw-2) or stave (UK)

> Staff is more common than stave in both [American English](https://en.wikipedia.org/wiki/American_English) and [British English](https://en.wikipedia.org/wiki/British_English).[[4]](https://en.wikipedia.org/wiki/Staff_(music)#cite_note-4) The plural of staff is staffs or staves in both American and British English.[[1]](https://en.wikipedia.org/wiki/Staff_(music)#cite_note-collins-1) (Stave is, in fact, a [back-formation](https://en.wikipedia.org/wiki/Back-formation) from staves.)

But yeah, either seems fine, not expecting another refactor either ;)

jaredjj3:
Haha, yeah, it's a confusing term. While the article disambiguated it for me, I ultimately chose "stave" to match `vexflow`. As you know, `musicxml` uses "staff", so there is still an implicit mapping, but I feel like that's at least neatly tucked away where it won't haunt me now :)

PR #132, opened 2023-10-07
Render lilypond03d

jaredjj3:
This PR renders 03d-Rhythm-DottedDurations-Factors.xml. It also adds a special case for rendering `<multiple-rest>` with a value of 1 by rendering a semibreve (whole) rest symbol.

## Divergences

- I speculate that GuitarPro 7 is inferring the rest representation from the note duration and divisions (more likely) OR using the multiple rest in combination with the time signature (less likely). It seems like there's a bug since it doesn't get the durations exactly to fill up the entire measure.
- `vexml` ignores the duration specifications and just shows a semibreve rest symbol for simplicity. I'm not sure what the exact implications of this are, since each would be 1 symbol. This might be an issue when creating an editor.
- `osmd` doesn't fork a special case when `<multiple-rest>` is 1.
- `lilypond` and MuseScore 4 both seem to render a breve rest symbol for measure 16. This is due to a conflict in the MusicXML. For measure 16, the `<multiple-rest>` value is 1, meaning it's just one measure being indicated as a rest. But the `<duration>` value 62 is equal to two full measures of silence in the given time signature. I decided that the `<multiple-rest>` takes precedence in `vexml`.

## vexml

image

## lilypond

image

## osmd

image

## MuseScore 4

image

## GuitarPro 7

image


jaredjj3:
I forgot to push my changes.

Issue #128, opened 2023-10-05
Handle mid-measure <attributes> changes 


jaredjj3:
## Problem

The [`<measure>`](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/measure-partwise/) element has the following children that we care about (so far):

- `<note>`
- `<backup>`
- `<forward>`
- `<attributes>`
- `<barline>`

Each of these elements can appear zero or more times.

In #120, we added the concept of [musicxml.MeasureEntry](https://github.com/stringsync/vexml/blob/44e8817a3be276b738ac7c1dcbc1de084d7c3ccd/src/musicxml/measure.ts#L9), which allowed us to correctly account for voice changes. However, in [03c-Rhythm-DivisionChange.xml](http://lilypond.org/doc/v2.25/input/regression/musicxml/collated-files#g_t03-_002e_002e_002e-rhythm), I learned that the `<attributes>` element can change mid measure.

Currently, there are the following concerns:

- A `rendering.Stave` is responsible for rendering `musicxml.MeasureEntry`.
- A `vexflow.Stave` is responsible for rendering a _single_ `vexflow.StaveModifier` and a _single_ group`vexflow.Tickable` objects tied to a `vexflow.Clef`.

Mid-measure `<attributes>` changes are at odds with the current rendering model. A `rendering.Stave` inherently doesn't support multiple `vexflow.Stave` objects to be rendered. This is very similar to the problem that #125 solved: a `rendering.Voice` was intended to only render a single `vexflow.Voice` object.

Ultimately, this seems like a modeling problem. Even though 03c-Rhythm-DivisionChange.xml renders correctly in `vexml`, we will inevitably run into issues when we want to create more dynamic features that rely on correct divisions accounting, such as a document cursor.

## Proposal

Create a `rendering.MeasureFragment` abstraction. A `rendering.Measure` will be composed of many `rendering.MeasureFragment` objects. The `rendering.Measure`'s responsibility will now become coordinating a number of fragments to render what _looks like_ a single cohesive measure. Under-the-hood, each measure will now be composed of several `vexflow.Stave` objects.

### Caveats

- The divisions accounting should probably still live in `rendering.Measure`, but I'm unsure if it will be possible to do so. If not, we can always have `rendering.Measure` provide the previous `rendering.MeasureEntry` to `rendering.MeasureEntry.create`, where there will be access to all the private properties without breaking encapsulation. 

[...]

Issue #122, opened 2023-10-03
MusicXML Player with vexml


infojunkie:
Like Pianoplay, I am also interested to use vexml to render MusicXML in my [player component](https://github.com/infojunkie/musicxml-player). Today, I support both OSMD and Verovio, and I find the latter more robust, although it is much harder to hack (C++ => wasm) so I would appreciate a simpler library. 

My basic requirements for integration:
- Ability to position cursor
- Ability to react on object (note) clicks / taps
- Ability to convert from measure number to timestamp and vice-versa

Happy to dig into this if there's interest from the maintainers here.

jaredjj3:
Thanks for reaching out. I haven't heard of Verovio until today. I read through it a bit, and I agree it's dense, but it's definitely a great choice especially if performance is the goal.

The requirements you gave seem reasonable. FWIW, I achieved those things on top of OSMD in stringsync.com. The library is [here](https://github.com/stringsync/stringsync/tree/master/web/src/lib/MusicDisplay). I plan to bring the lessons learned from that project to `vexml`. Let me know if you have questions.

>Happy to dig into this if there's interest from the maintainers here.

At the moment, I consider the API and inner-workings unstable. Would you mind holding off until the library can successfully render full scores reliably?

Right now we're nailing smaller components in the [lilypond test suite](https://lilypond.org/doc/v2.23/input/regression/musicxml/collated-files.html). After that, we'll see how well we can render scores, which will inform how ready we are to work on more dynamic features.

infojunkie:
Thanks for your reply. Sure, I'll keep and eye here and wait for a release.

PR #135, opened 2023-10-08
Render lilypond11a

jaredjj3:
This PR renders 11a-TimeSignatures.xml by adding support for more advanced time signatures. This PR does **not** add support for note ornaments such as slurs. It changes the default time signature from 4/4 to common time.

## Divergences

- GuitarPro 7 renders the number-based time signatures instead of the symbols.
- `vexml` does not render slurs yet (measure 8).
- MuseScore 4 and `lilypond` render measure fragments with a time signature change but no notes. I decided that this looks too disparate.

## vexml

image

## lilypond

image

## osmd

image

## MuseScore 4

image

## GuitarPro 7

image


rvilarl:
@jaredjj3 thanks for the amazing progress. I am looking forward to making a pianoplay branch.

jaredjj3:
Sure thing! Once I had a convincing maintenance story, I finally felt good enough to run with it. I knew this was going to be a very complex library, which is why that was so important to me.

Like I mentioned in #122, I recommend waiting to implement dynamic features until the [*Rendering](https://github.com/stringsync/vexml/blob/422ac560e0b454365bbecd3f246667e0f822d2e5/src/rendering/score.ts#L14) 
 types are stable. I'm not convinced that I've surfaced all the incorrect assumptions I've made about MusicXML / vexflow / Music Notation and I might have to make some ~major structural changes like in #128 .

PR #139, opened 2023-10-08
Render lilypond11d

jaredjj3:
This PR renders 11d-TimeSignatures-CompoundMultiple.xml. 

>**NOTE:** `vexflow` doesn't seem to support rendering compound time signatures, but this case seems to exceedingly rare, the fallback behavior is acceptable.

## Divergences

- `vexml` doesn't support beaming yet.
- `vexml` and `osmd` don't support multiple complex time signatures, but reduce to the same "effective" time signature using the LCM of the denominators.
- MuseScore 4 seems to only use the _last_ time signature in the `<time>`.
- GuitarPro 7 seems to only use the _first_ time signature in the `<time>`.

## vexml

image

## lilypond

image

## osmd

image

## MuseScore 4

image

## GuitarPro 7

image


rvilarl:
@jaredjj3 vexml supports compound time signatures. Like this:

```js
new Stave(10, 10, 300)
    .addTimeSignature('3/8')
    .addTimeSignature('+')
    .addTimeSignature('2/8')
    .addTimeSignature('+')
    .addTimeSignature('3/4');
```

jaredjj3:
Awesome! I’m glad you saw this. I’ll fix this.

jaredjj3:
I was able to use the fix in #143, thanks again @rvilarl.

PR #142, opened 2023-10-08
Render lilypond11g

jaredjj3:
This PR renders 11g-TimeSignatures-SingleNumber.xml. It also adds support for `single-number` time signatures, which updated the snapshot for 11f-TimeSignatures-SymbolMeaning.xml.

## Divergences

The renderers here don't support this kind of time signature. `vexflow` doesn't support this either, but I think this is a good signal to accept this divergence.

## vexml

image

## lilypond

image

## osmd

image

## MuseScore 4

image

## GuitarPro 7

image


rvilarl:
@jaredjj3 did you try?
```js
new Stave(10, 10, 300)
    .addTimeSignature('3');
  ```

jaredjj3:
Hmm I thought I did. I’ll try again when I resume in a few hours or tomorrow.

jaredjj3:
@rvilarl, this doesn't seem to work: https://jsfiddle.net/gynua7pv/3/. Am I doing something wrong?

rvilarl:
I will look into that, it seems to be a vexflow problem.

PR #146, opened 2023-10-10
Render lilypond12a

jaredjj3:
This PR renders 12a-Clefs.xml.

## Divergences

- `vexml` and `osmd` seem to render notes in the wrong place, this seems like a problem with `vexflow`.
- `vexml` and `osmd` don't render the "8" symbol on some of the uncommon clefs (example, measure 5).
- `vexml`'s tablture rendering is terrible, but I'm willing to accept that until the lilypond71* tests. This is also an exceptional case, rendering tablature _inline_ of other clef types.
- MuseScore 4 and GuitarPro 7 seem to ignore the TAB rendering.

## vexml

image

## lilypond

image

## osmd

image

## MuseScore 4

image

## GuitarPro 7

image


jaredjj3:
This one is taking a little longer than expected because of the following convention:

>If a clef changes at a measure boundary, the barline should be rendered **after** the clef symbol.

I really want to avoid making rendering objects "looking" ahead and I want to avoid mutations outside of the `create` static factory methods. The solution I'm landing on is:

- Create the concept of `StaveSignature`, which is a wrapper around `<attributes>` that also accounts for changes throughout the score. The crucial part is that `nil` values actually mean "use the accumulated value so far".
- Create a `StaveSignatureRegistry`, which is an index of `[measureIndex, measureEntryIndex] -> StaveSignature`. As long as callers can construct a key correctly, they can get the `StaveSignature` objects they're looking for. This includes previous and next `StaveSignature` objects, allowing callers to effectively "look ahead" without explicitly exposing the next entire  rendering object.

jaredjj3:
Saving my thoughts somewhere so I don't forget.

I failed to develop a purely functional solution for my barline issues. I thought I could get away with "looking ahead", but it proved to be much more difficult as it also depends on the measure's position in the system, which isn't known at `create` time.

I don't think I can maintainably push all the accounting aspects down the `create` hierarchy, because then I'd need to recalculate everything when I eventually start slicing a `rendering.Part` post `create`.

I think ultimately `System.split` needs to perform mutations on the resulting systems:

- Iterate over the resulting `MeasureFragment`
- For each `MeasureFragment`, look at its _next_ `MeasureFragment`.
  - If the next `MeasureFragment` is on the same system and contains a clef change _across a measure boundary_, then mutate the next `MeasureFragment`'s `endBarStyle` to be what your `endBarStyle` would've been.
  
I think this algorithm will break if this needs to be done for more than one `MeasureFragment` consecutively, so I need to maintain what to original end bar style was going to be at all times.

Maybe `MeasureFragment` should break it's endBarStyle into two sub properties:

```ts
type TentativeProperty<T> = {
  initial: T;
  actual: T | null;
};
```

The mutation would propagate from the _previous_ `MeasureFragment.endBarStyle.initial` to the _next_ `MeasureFragment.endBarStyle.actual`.

jaredjj3:
@rvilarl, you might find these results interesting, since both `vexml` and `osmd` had some incorrect (double check me) note placements.

- [clef type calculation](https://github.com/stringsync/vexml/blob/219e11f8143f26b5bf5e95e04e926d267d24c3b6/src/musicxml/clef.ts#L33)
- [clef type comsumption](https://github.com/stringsync/vexml/blob/219e11f8143f26b5bf5e95e04e926d267d24c3b6/src/rendering/note.ts#L148)

jaredjj3:
I just discovered [Stave.addEndClef](https://github.com/0xfe/vexflow/blob/7e7eb97bf1580a31171302b3bd8165f057b692ba/src/stave.ts#L562). I think there's an opportunity to simplify the [measure fragment creation](https://github.com/stringsync/vexml/blob/a5aee8445be5cd5f38fdbb6368677de213a3e426/src/rendering/measure.ts#L135-L158) and [measure fragment padding](https://github.com/stringsync/vexml/blob/a5aee8445be5cd5f38fdbb6368677de213a3e426/src/rendering/measurefragment.ts#L117) calculations.

rvilarl:
@jaredjj3 you seem not to be using the annotation field to get the 8 on the clefs:

stave.addClef('treble', 'default', **'8va'**)

jaredjj3:
Ah, thanks! Seeing how that's an octave sign (?), that could explain why the notes were placed wrong. I'll fix that in a future PR.

jaredjj3:
> I just discovered [Stave.addEndClef](https://github.com/0xfe/vexflow/blob/7e7eb97bf1580a31171302b3bd8165f057b692ba/src/stave.ts#L562). I think there's an opportunity to simplify the [measure fragment creation](https://github.com/stringsync/vexml/blob/a5aee8445be5cd5f38fdbb6368677de213a3e426/src/rendering/measure.ts#L135-L158) and [measure fragment padding](https://github.com/stringsync/vexml/blob/a5aee8445be5cd5f38fdbb6368677de213a3e426/src/rendering/measurefragment.ts#L117) calculations.

It turns out the solution I have is elegant, because I don't have to change how widths are accommodated. I'm leveraging the existing `MeasureFragment` machinery to do everything.

jaredjj3:
> @jaredjj3 you seem not to be using the annotation field to get the 8 on the clefs:
> 
> stave.addClef('treble', 'default', **'8va'**)

This worked. I'm going to push a PR now.

PR #148, opened 2023-10-14
Render lilypond13a

jaredjj3:
This PR renders 13a-KeySignatures.xml.

## Divergences

- `vexml` doesn't render cancel keys.
- `lilypond` renders some unnecessary cancel keys.
- `vexml` renders alterations in the circle of fifths order while `lilypond` seems to append at the end. I prefer `vexml`.
- `lilypond`, MuseScore 4 and Guitar Pro 7 render accidentals for some reason. There are no specified `<accidental>` elements in 13a-KeySignatures.xml.
- MuseScore 4 and Guitar Pro 7 don't seem to support these non-western key signatures.

## vexml

image

## lilypond

image

## osmd

Does not render.

## MuseScore 4

image

## Guitar Pro 7

image


jaredjj3:
Breadcrumbs: [Vexflow altered key examples](https://github.com/0xfe/vexflow/blob/7e7eb97bf1580a31171302b3bd8165f057b692ba/tests/keysignature_tests.ts#L210)

jaredjj3:
>`lilypond`, MuseScore 4 and Guitar Pro 7 render accidentals for some reason. There are no specified `<accidental>` elements in 13a-KeySignatures.xml.

It just dawned on me... The reason why there are natural symbols is because the MusicXML document specifies the following for all `<note>` elements:

```xml
<pitch>
  <step>C</step>
  <octave>4</octave>
</pitch>
```

In order to truly honor the pitch, we need to use a natural accidental. This is likely going to be complex, especially considering microtonal music. I'm going to look at this now.

jaredjj3:
Also, it's worth mentioning that cancel keys are definitely necessary when going _to_ C major / A minor. Otherwise, the musician has no indication that the time signature changed.

PR #149, opened 2023-10-15
Fix lilypond12a rendering

jaredjj3:
This PR fixes the issues introduced in #146.

- When rendering clefs, `vexml` now checks the octave change.
- When calculating `StaveNote` keys, `vexml` now correctly accounts for the clef's octave change.

## Divergences

The main issue I see is how percussion notes are rendered in `vexml` vs `lilypond`. Since percussive sounds technically have a pitch, I think the right thing to do is to honor the `<pitch>` specification and not default to rendering on the same line as the percussion clef.

## vexml

image

## lilypond

image

## osmd

image

## MuseScore 4

image

## GuitarPro 7

image


rvilarl:
@jaredjj3 great! With TAB  it looks as a measure and a note was missing. Are the TAB and the treble on different ones?
@sschmidTU it seems tha OSMD is not presenting the notes right with this test.

jaredjj3:
Thanks! Yeah for the TAB measure in the MusicXML, there is a note without any position information attached to it, which is why there’s weird behavior.

In vexml, I decided to ignore it for now. osmd seems to be defaulting it to 0 on the 1st string (highest stave line).

jaredjj3:
>@sschmidTU it seems that OSMD is not presenting the notes right with this test. The TAB is also missing.

FWIW, I got the same result when I added the clef octave change to the note octave. I fixed it by subtracting the clef octave change from the note octave.

sschmidTU:
The OSMD issue with 8va clef measures sometimes not having the right note line is weird.
I tested it with two simpler samples where it doesn't appear, there's no inherent issue with switching to an 8va clef.
image

image

[test_clef_octaveshift_2_samples.zip](https://github.com/stringsync/vexml/files/12916876/test_clef_octaveshift_2_samples.zip)

This is the same rendering as in MuseScore.

I can only assume that the academic and complex nature of the lilypond sample and maybe encoding causes some rare issue here.

It's also hard to debug.

If this doesn't happen in any practical samples and is hard to debug, I would assign this low priority in OSMD.

The peak of this academicness by the way is the switch to a TAB clef between measures in my opinion. That's absolutely never going to happen. It also doesn't make sense, because these are two different types of notation with different stafflines.
I wouldn't waste a second on this. Even lilypond doesn't display it correctly.
image

Some test samples like in lilypond have this tendency to be academic. And while I'm all for maximizing correctness and test coverage, music rendering is a never ending trove of possible optimizations and complications, so you have to go by priority.

Still, thanks to both of you for the hints and pings, sincerely. It could have been that this was a ubiquitous issue in OSMD and easy to fix. I'm generally enjoying looking at these tests and seeing the differences between the renderers.

sschmidTU:
By the way, the reason OSMD doesn't use the lilypond test suite is that we already have our own continually expanding test suite, and there would be a huge amount of overlap between both suites in terms of what is tested.
And our visual regression tests already take around 2 minutes 15 seconds for 198 samples on my 2022 machine, so I would estimate the 127 lilypond samples to add almost 2 minutes to that.

Also, the Lilypond suite are very old, so I'm not sure if they even test any MusicXML 3 or 4 features.
They also call the files `.xml`, while the [musicxml file extension](https://www.loc.gov/preservation/digital/formats/fdd/fdd_explanation.shtml#sign) has been [recommended to be .musicxml since MusicXML 3.1 in 2017](https://www.loc.gov/preservation/digital/formats/fdd/fdd000499.shtml), while .xml is still accepted.

jaredjj3:
@sschmidTU, I totally agree with your approach. I'm starting to hit the point in the test suite where it's purely academic and I think prioritization is more important than complete coverage. Something something [pareto principle](https://en.wikipedia.org/wiki/Pareto_principle).

For example, I'd rather not add  `<key-step>` and `<key-alter>` support for something that I suspect will be rarely used. These are next up, but I'm going to defer supporting these.

image

>Also, the Lilypond suite is very old, so I'm not sure if they even test any MusicXML 3 or 4 features.

Yeah, I suspect that if we were to _solely_ rely on lilypond, we'd miss coverage for the newest features. However, I can say with confidence that they are indeed MusicXML 4 compatible, which was the intent as MusicXML major versions advanced. In [@stringsync/musicxml](https://github.com/stringsync/musicxml/tree/master), I wrote [compatibility tests](https://github.com/stringsync/musicxml/tree/master/src/compatibility) for lilypond and OSMD.

All in all, now that I'm getting to the more academic test cases for lilypond, I'm going to start considering other test suites centered around practicality. Thanks for sharing your perspective and keeping me honest :)

PR #151, opened 2023-10-15
Move TimeSignature to rendering

jaredjj3:
This PR simply moves `TimeSignature` from `musicxml` to `rendering` in order to be more aligned with the intent of `musicxml` being a pure data library.


jaredjj3:
Draft: I need to write tests for the new `musicxml.Time` API.

PR #152, opened 2023-10-15
Fix lilypond13a

jaredjj3:
This PR fixes #148.

- Cancels the previous stave's key signature.
- Renders accidentals when there's either an explicit `<accidental>` element or the `<pitch>` does not match the key signature.

## Divergences

Some of the cancel keys don't render nicely in `vexml`. Other than that, the differences are unremarkable.

## vexml

13a-KeySignatures_900px

## lilypond

image


rvilarl:
> Some of the cancel keys don't render nicely in vexml

@jaredjj3 what do you mean?

jaredjj3:
In measures 13, 15, 19, and 21, the natural sign seems too close to the flats.

rvilarl:
you are right

PR #154, opened 2023-10-16
Fix first measure stave signature bug

jaredjj3:
This PR fixes a bug where the first stave signature was never being properly considered. This happened when partitioning an array of `MeasureEntry` objects based on `StaveSignature`. We almost always erroneously assigned the _second_ stave signature to the very first measure fragment.


src/rendering/measure.ts
```diff
@@ -144,10 +144,8 @@ export class Measure {
       const isLastMeasureEntry = measureEntryIndex === opts.measureEntries.length - 1;
 
       if (measureEntry instanceof StaveSignature) {
-        staveSignature = measureEntry;
```
jaredjj3, line 147: The bug was that we reassigned `staveSignature` before creating the next `MeasureFragment` with it below.

PR #157, opened 2023-10-17
Render lilypond31c

### LINKED ISSUE #62: Missing: Metronome marks

jaredjj3:
This PR fixes #62 and renders 31c-MetronomeMarks.xml.

## Divergences

- `vexml`, `osmd`, MuseScore 4, and Guitar Pro 7 don't support all metronome marks.
- The note spacing in `vexml` is dependent on the metronome marks while the other renders are independent.
- `vexml`, `osmd`, and Guitar Pro 7 don't render parentheses.
- MuseScore 4 renders metronome marks below the stave.

## vexml

image

## lilypond

image

## osmd

image

## MuseScore 4

image

## Guitar Pro 7

image


jaredjj3:
It looks like [`vexflow.Stave.setTempo`](https://github.com/0xfe/vexflow/blob/7e7eb97bf1580a31171302b3bd8165f057b692ba/src/stave.ts#L325) is the `vexflow` API I need to hook into.

The `<direction>` element contains the metronome information:

```
<direction>
  <direction-type>
    <metronome>
      <beat-unit>quarter</beat-unit>
      <beat-unit-dot/>
      <per-minute>100</per-minute>
    </metronome>
  </direction-type>
</direction>
```

Based on the [MusicXML spec](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/measure-partwise/), `<measure>` can have zero or more `<direction>` elements, similar to `<attributes>`. This is evident in 31c-MetronomeMarks.xml.

I'm pretty sure I can leverage `vexml`'s measure fragment abstraction to render several metronome marks per measure.

**Action Items**

- Add `musicxml` support for `<direction>`.
- Add `musicxml` support for `<direction-type>`. This element has a lot of variations on what it could be. Include a helper that tells the caller what type it is.
- Create a `Metronome` rendering object.
- Add `Metronome` as a child of `MeasureFragment`
- Update the measure fragmenting procedure to create a new fragment for every `Metronome`.
- Supply the `Stave` to the `Metronome` at `MeasureFragment` render time. Although, it makes me wonder if `Metronome` should actually be a child of `Stave`. Consider this alternative a little more carefully.

>**NOTE:** I assume that if the MusicXML writer included the `<direction>`, they're expressing that they want it rendered at a specific place regardless of what `Metronome` is "active". I really don't think this should be handled like `StaveSignature`/`<attributes>`, but I need to think about this more carefully.

rvilarl:
Hi @jaredjj3 anything to implement on the vexflow side? The parenthesis? One question, why are there so many "120" compared to the others?

jaredjj3:
>Hi @jaredjj3 anything to implement on the vexflow side? The parenthesis?

`vexflow` only seems to support numerical `bpm` specifications and does not support parentheses ([vexflow.StaveTempoOptions](https://github.com/0xfe/vexflow/blob/7e7eb97bf1580a31171302b3bd8165f057b692ba/src/stavetempo.ts#L12)). This seems pretty exceptional, so I would put the feature request at a very low priority, especially considering that it would be low priority to implement in `vexml`, too.

>One question, why are there so many "120" compared to the others?

It was a default doing more harm than good. I fixed this in #160.

PR #162, opened 2023-10-29
Render lilypond33a

### LINKED ISSUE #63: Missing: Pedals

jaredjj3:
This PR fixes #63.

This probably should have been broken into smaller test cases. However, being exposed to many different spanners ended up being a good thing. I initially thought that the different spanners in MusicXML can be modeled using some common abstraction, but I was wrong. The way the different spanners express `start`, `continue`, and `stop` vary.

I also found it very difficult to find a rendering vehicle for `<direction-type>` elements. As stated in the [docs](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/direction/):

>A direction is a musical indication that is not necessarily attached to a specific note.

However, `vexflow` usually uses `Note` or `StaveNote` as anchors for rendering various spanner symbols. Therefore, I had to make the association myself in `vexml`. In contrast, `<ornaments>` are nested under `<note>` elements, which made the association straightforward.

I decided to ignore rendering `<bracket>` and `<dashes>` since they don't seem to be supported in `vexflow`, and only MuseScore 4 does a good job of rendering them.

Ultimately, I think this PR has gone on long enough and I should submit it as a checkpoint and work on other things for the time being. I'm not totally satisfied with the trajectory of the spanner implementation in `vexml`, and I left several TODOs to revisit later.

The main thing that I dislike is that the `rendering.*` spanner objects need something already rendered in order to be _created_. However, they inherently require you to look ahead when reading data (so you can find when they `stop`) or you need to defer their calculation until you know everything about it. This seems to call for the builder pattern, but I'm not going to explore it right now.

## Divergences

- `lilypond`, MuseScore 4, and Guitar Pro 7 don't honor the `tuplet[placement="above"]` attribute in measure 1.
- `vexml`, `osmd`, and Guitar Pro 7 don't honor `slur[line-type="dashed"]` in measure 3.
- `vexml` renders trill marks awkwardly in measure 6 and 7. `lilypond` doesn't render trill marks. `osmd` renders trill symbols but no vibrato brackets.
- All renderers handle octave shifts correctly.
- Only MuseScore 4 renders `<bracket>` and `<dashes>`.
- `vexml`, `osmd`, and Guitar Pro 7 don't render `<glissando>` elements in measure 16.
- Only `lilypond` renders `<bend-alter>` in measure 17. This makes me think that the MusicXML might be wrong.
- Only `lilypond` and `osmd` render `<slide>` elements in measure 18. I expected Guitar Pro 7 to handle this, since slides are common with stringed instruments.
- Only `lilypond` and Guitar Pro 7 render `<tremolo>` symbols in measure 20. MuseScore 4 uses a more spanner-like symbol, but has these unusual eight rests rendered between notes.
- `vexml` and `osmd` default to rendering bracket pedal markings in measure 23. MuseScore 4 uses a `#` symbol to indicate `pedal[type="change"]` (not sure if this is valid).

## vexml

image

## lilypond

image

## osmd

image

## MuseScore 4

image

## Guitar Pro 7

image


jaredjj3:
I need to add [`<notations>`](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/notations/) support to `vexml`. However, it doesn't seem to have meaningful attributes other than `print-object`, and I assume that `[print-object="no"]`  is very rare.

I think it makes sense for me to make `musicxml.Note` bypass `<notation>`, and query the things we care about directly. However, it seems like `musicxml.Note` already has a lot attached to it.

Thinking more about provenance, I think it's better to have a more granular abstraction.

---

The plan is to create a new `musicxml.Notation` object that will how all of the different notation types inside of it. The tradeoff is more complex code dealing with the abstraction, but the main pro is that it's less obfuscated from the MusicXML spec.

jaredjj3:
Breadcrumbs:

- [vexflow Tuplet test results](https://vexflow.com/tests/index.html?module=Tuplet)
- [vexflow Tuplet test code](https://github.com/0xfe/vexflow/blob/7e7eb97bf1580a31171302b3bd8165f057b692ba/tests/tuplet_tests.ts)

jaredjj3:
The `vexflow` tuplet API is pretty straight forward:

[source](https://github.com/0xfe/vexflow/blob/7e7eb97bf1580a31171302b3bd8165f057b692ba/tests/tuplet_tests.ts#L61C42-L61C42)

```ts
f.Tuplet({ notes: notes.slice(0, 3) });
f.Tuplet({ notes: notes.slice(3, 6) });
```

It seems that we need `vexflow` notes, so I plan to follow the same [strategy](https://github.com/stringsync/vexml/blob/3ccb1228cbafb45f7086bfd1e504425ff96af817/src/rendering/measurefragment.ts#L138) as how `vexml` renders beams.

1. Render the `rendering.Note` to get the `vexflow.Tickable` (not sure on the exact type here). The `NoteRendering` should also include the tuplet data.
2. Use the tuplet data to construct the `vexflow.Tuplet` objects.
3. In `rendering.Score`, draw them after the `vexflow.Tickable` objects have been drawn.

rvilarl:
> Breadcrumbs:
> 
> - [vexflow Tuplet test results](https://vexflow.com/tests/index.html?module=Tuplet)
> - [vexflow Tuplet test code](https://github.com/0xfe/vexflow/blob/7e7eb97bf1580a31171302b3bd8165f057b692ba/tests/tuplet_tests.ts)

Hi Jared,

Which version of Vexflow is this one? Do you know?

Regards,

Rodrigo

jaredjj3:
Thanks for calling this out, looking at https://vexflow.com/build/vexflow-tests.js, I see that I'm running

>VexFlow 1.2.77 built on 2016-08-21.

Are there plans to update vexflow.com/tests?

rvilarl:
I will setup an updated one. But I am in Brussels and I forgot my computer charger at home. :( I will be back on Wednesday

jaredjj3:
> I will setup an updated one. But I am in Brussels and I forgot my computer charger at home. :( I will be back on Wednesday

No rush on my end — I can always run locally. Thanks for always double checking me. Enjoy Brussels!

jaredjj3:
https://github.com/stringsync/vexml/pull/162#issuecomment-1784099989

This strategy worked. I was hoping to see some other measures at least partially rendered, but they must be using some other MusicXML elements that are not covered yet.

image

jaredjj3:
I briefly studied slurs and ties to make sure I don't miss an opportunity to handle both at the same time. The TL;DR is that I can't and shouldn't — they use different MusicXML structures.

- slur: Uses the [`<slur>`](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/slur/) element. Uses a `start-stop-continue` like [`<beam>`](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/beam/), so I'll follow the same strategy as https://github.com/stringsync/vexml/pull/162#issuecomment-1784099989.

- tie: Uses the [`<tie>`](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/tie/) element. This spans across measures, which implies that it'll also span across measure fragments. I think I need to treat this like a beam, but at the measure level.

rvilarl:
You also have TextBrackets there.  😀

sschmidTU:
> I assume that `[print-object="no"]` is very rare.

Note: This is not that rare. 866 usages in the OSMD test samples.

image

The biggest use case is invisible notes. 

image

If you render notes that have print-object="no" (supposed to be invisible), you'll get this:
image

Often when writing scores in a sheet music program, the program basically forces you to add these rests for sake of correctness, but you can set them to be invisible, which is a common practice. You can also sometimes outright delete them, but that's a less common practice afaik.

This is also quite common for rests:
Respecting print-object="no":
image

Ignoring print-object="no":
image
(Schubert - An die Musik mm.19)

There are even better and more common examples where there are two rests in one measure for 2 voices, but you only need to show one common rest. "Unfortunately" OSMD does that anyways, even when ignoring the attribute.
image
(Land der Berge, end of measure 4)

@jaredjj3 You don't need to support this for all elements immediately, but for notes, I think it's quite important. Other interesting cases are key signature and time signature (not rendering them).

sschmidTU:
Oh and a simple trick to simplify the implementation is to render the notes with invisible color (alpha). A benefit there is that you can instantly (low performance impact) make them visible again during runtime by changing the alpha/color in the SVG, no need to render the whole sheet again.

jaredjj3:
@sschmidTU thanks for taking the time to write this. The edge cases you provided are illuminating and is preventing the project a lot of pain that would've been felt without your involvement.

>You can also sometimes outright delete them, but that's a less common practice afaik.

The fundamental I've been trying to adopt is "respect the engraver's intentions as long as vexflow permits it". I really like your approach of making notes invisible instead because it honors the musicxml, doesn't alter the structure, and offers a performant solution to toggle on-and-off.

I'm on mobile at the moment, but I'll definitely digest your other hints as I continue implementing this.

rvilarl:
> Thanks for calling this out, looking at https://vexflow.com/build/vexflow-tests.js, I see that I'm running
> 
> > VexFlow 1.2.77 built on 2016-08-21.
> 
> Are there plans to update vexflow.com/tests?

Now available in https://rvilarl.github.io/vexflow-examples/tests/flow.html

Build in progress for examples, tutorial... https://rvilarl.github.io/vexflow-examples/

jaredjj3:
@rvilarl, two questions.

1. Does https://rvilarl.github.io/vexflow-examples/tests/flow.html load for you? The page seems to freeze up, but I haven't waited more than a minute.
2. I added supported slurs, but I found that I couldn't make them dashed to support the [`slur[line-type]`](https://www.w3.org/2021/06/musicxml40/musicxml-reference/data-types/line-type/) property. Is this supported in vexflow?

rvilarl:
> @rvilarl, two questions.
> 
> 1. Does https://rvilarl.github.io/vexflow-examples/tests/flow.html load for you? The page seems to freeze up, but I haven't waited more than a minute.

yes it does but it takes almost two minutes

> 2. I added supported slurs, but I found that I couldn't make them dashed to support the [`slur[line-type]`](https://www.w3.org/2021/06/musicxml40/musicxml-reference/data-types/line-type/) property. Is this supported in vexflow?

I think not. Would we need it on curves, grace notes or both?

jaredjj3:
@rvilarl I'm not sure. You can scan the [lilypond test suite](https://lilypond.org/doc/v2.23/input/regression/musicxml/collated-files.html) or just wait for me to mention you for missing `vexflow` features again. It's probably better to wait until we fully understand.

jaredjj3:
@rvilarl, I'm currently stuck on rendering crescendos. Looking at the vexflow [tests](https://github.com/0xfe/vexflow/blob/7e7eb97bf1580a31171302b3bd8165f057b692ba/tests/textnote_tests.ts#L230), it seems that these are modeled as `Note` objects, which implies they have a duration.

However, this seems to break down when they span multiple measures. How is the `Crescendo` object supposed to know anything about the measures after it?

I guess what I'm asking is: **How do I render crescendos across multiple measures?**

@sschmidTU is this something you patched into vexflow in osmd? I tested osmd, and it works beautifully.

image

rvilarl:
> @rvilarl, I'm currently stuck on rendering crescendos. Looking at the vexflow [tests](https://github.com/0xfe/vexflow/blob/7e7eb97bf1580a31171302b3bd8165f057b692ba/tests/textnote_tests.ts#L230), it seems that these are modeled as `Note` objects, which implies they have a duration.
> 
> However, this seems to break down when they span multiple measures. How is the `Crescendo` object supposed to know anything about the measures after it?
> 
> I guess what I'm asking is: **How do I render crescendos across multiple measures?**
> 

You have to use StaveHairpin, look at https://rvilarl.github.io/vexflow-examples/examples/index.html

jaredjj3:
Ah, thank you. This example is perfect and I'm unblocked now.

There is an interesting edge case where a spanner crosses system (not measure) boundaries. I need to make sure I restart a spanner correctly on the new system when this happens.

So far I can only think of ties and crescendos having this issue. The other types of spanners seem that they would be contained within the measure and therefore not span systems.

sschmidTU:
@jaredjj3 The OSMD question is now obsolete since you know about StaveHairpin now, right? By the way, OSMD calculates and renders the wedges / hairpins / (de-)crescendo lines itself by drawing two lines. There are some things in OSMD we don't do via Vexflow but draw ourselves.

jaredjj3:
@sschmidTU ahhh that makes sense. I think I saw that when grokking osmd and was wondering what vexflow machinery you were using. It turns out that you have your own layout engine.

As I add more features, I think I might have to roll out something similar to unify lyrics and ornaments across measures. Otherwise, I'm not sure how `vexflow.Stave` are supposed to inherently know about other `vexflow.Stave`, even when using `vexflow.System`.

As always, thanks for the tip!

rvilarl:
@jaredjj3 could you update the TODO in the first comment to see how far you are? 

Apart from the dotted slur and taking into accoutn your comment above would it help to have a score class to VexFlow?

a score could contain several systems and could take the responsability, for example, to place them one after another. We could also consider other functionality at score level.

jaredjj3:
>@jaredjj3 could you update the TODO in the first comment to see how far you are?

Done. I refactored a good amount of code around and I've been somewhat busy these past few weeks, so I haven't made a lot of new progress. Luckily, the rest of the spanners seem to ~mostly follow the same patterns I'm using for beams, tuplets, and slurs, so I can pretty much copy the approach. I also looked up the vexflow APIs for the rest of the spanners and I was only confused with crescendo.

I should be able to finish this weekend :)

jaredjj3:
> would it help to have a score class to VexFlow?
>
>a score could contain several systems and could take the responsability, for example, to place them one after another. We could also consider other functionality at score level.

I would like to think about this carefully. Do you have a sense how other people are addressing the problem of orchestrating symbols across measures that should be on the same horizontal line?

rvilarl:
> > would it help to have a score class to VexFlow?
> > a score could contain several systems and could take the responsability, for example, to place them one after another. We could also consider other functionality at score level.
> 
> I would like to think about this carefully. Do you have a sense how other people are addressing the problem of orchestrating symbols across measures that should be on the same horizontal line?

I agree that we have to think it carefully. And answering to your question I do not know. :)

jaredjj3:
Let's move the discussion to vexflow and try to come up with a few canonical examples demonstrating this problem. I'll make the issue when I have time, but feel free to make it if you're able to do it sooner than me.

rvilarl:
> Let's move the discussion to vexflow and try to come up with a few canonical examples demonstrating this problem. I'll make the issue when I have time, but feel free to make it if you're able to do it sooner than me.

I made it yesterday https://github.com/vexflow/vexflow/issues/158

rvilarl:
@jaredjj3 why didn't you use the StaveHairpin for the wedges?

jaredjj3:
@rvilarl, I think I'd the trill marks should render above the stave, but maybe this is expected behavior. Would you confirm please?

I used [this ornament test](https://github.com/0xfe/vexflow/blob/7e7eb97bf1580a31171302b3bd8165f057b692ba/tests/ornament_tests.ts#L66) as a vexflow integration guide.

image

rvilarl:
@mscuthbert where should be the tr placed?

@mscuthbert @ronyeh do you use a paid software to compare? 
Ron normally you refer to Dorico documentation like in https://github.com/vexflow/vexflow/pull/160#issuecomment-1806990103

jaredjj3:
@rvilarl, as move forward with the implementation, I'm noticing some other spanners that don't translate well to vexflow. It could just be that I'm using the library improperly. I also think the musicxml from the lilypond test suite might be wrong or outdated. I need to find some canonical examples of `<wavy-line>` usages before locking into an implementation.

See measures 6 and 7. MuseScore4 does the best job of rendering this, as the wavy lines are a continuum instead of discrete symbols tied to notes. Is there something like `vexflow.StaveHairpin` that allows me to represent the wavy line as a spanner that goes from one note to another?

image

rvilarl:
Hi @jaredjj3 I think you need VibratoBrackets

https://vexflow.github.io/vexflow-examples/tests/flow.html?moduleId=b23f2b03

jaredjj3:
@rvilarl, thanks this fixed it.

image

ronyeh:
> do you use a paid software to compare?

No I don't use a paid music engraver. I mention dorico a lot because it is made by Steinberg, the same company that made Bravura and Petaluma fonts. 

I'd be happy if vexflow generates output comparable to lilypond or musescore.

rvilarl:
> > do you use a paid software to compare?
> 
> No I don't use a paid music engraver. I mention dorico a lot because it is made by Steinberg, the same company that made Bravura and Petaluma fonts.
> 
> I'd be happy if vexflow generates output comparable to lilypond or musescore.

I just saw that there is a free version that imports musicxml. But only for mac and windows. I use Linux :(. @jaredjj3 what SO do you use? Perhaps you can use it for your comparations. 

https://www.steinberg.net/dorico/se/

rvilarl:
> @rvilarl, I think I'd the trill marks should render above the stave, but maybe this is expected behavior. Would you confirm please?

@jaredjj3 I think you are right. I will fix it. By the way shaded slurs are already in, we will make a new release soon so that you can include it.  :)

ronyeh:
> But only for mac and windows. I use Linux

Are you able to run it via Wine?

jaredjj3:
> > > do you use a paid software to compare?
> > 
> > 
> > No I don't use a paid music engraver. I mention dorico a lot because it is made by Steinberg, the same company that made Bravura and Petaluma fonts.
> > I'd be happy if vexflow generates output comparable to lilypond or musescore.
> 
> I just saw that there is a free version that imports musicxml. But only for mac and windows. I use Linux :(. @jaredjj3 what SO do you use? Perhaps you can use it for your comparations.
> 
> https://www.steinberg.net/dorico/se/

I'm not sure what SO is, but I'm on macOS. I'll take a look at dorico sometime. It would be good for me to get another comparison :)

rvilarl:
> > > > do you use a paid software to compare?
> > > 
> > > 
> > > No I don't use a paid music engraver. I mention dorico a lot because it is made by Steinberg, the same company that made Bravura and Petaluma fonts.
> > > I'd be happy if vexflow generates output comparable to lilypond or musescore.
> > 
> > 
> > I just saw that there is a free version that imports musicxml. But only for mac and windows. I use Linux :(. @jaredjj3 what SO do you use? Perhaps you can use it for your comparations.
> > https://www.steinberg.net/dorico/se/
> 
> I'm not sure what SO is, but I'm on macOS. I'll take a look at dorico sometime. It would be good for me to get another comparison :)

SO is OS in Spanish: Sistema Operativo vs Operating System. :)

jaredjj3:
I have a naive implementation of spanners, but I'd like to revisit and think of a more maintainable approach. @rvilarl, you can test out the different [`vexflow.PedalMarking.type`](https://github.com/0xfe/vexflow/blob/7e7eb97bf1580a31171302b3bd8165f057b692ba/src/pedalmarking.ts#L75) by varying `pedal[line]` and `pedal[sign]` ([docs](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/pedal/)) in MusicXML. This test only shows the default.

Feel free to open more issues regarding pedal marks. I'm especially curious how this behaves when a pedal mark spans to another system.

rvilarl:
@jaredjj3 did you ignore the other ornaments and tremolos?
By the way the tr will go above the stave on the next release https://github.com/vexflow/vexflow/pull/165

jaredjj3:
Yes, I ignored tremolo because I couldn't find how to do it in the `vexflow` tests right away. I also noticed different representations, so I wanted to see how other software renders it because committing to a certain path. I'll do this in a downstream PR.

Glad to hear about the `tr` symbol! I'm not sure how the `tr` symbol and the `VibratoBracket` interact, but I think ideally it looks like MuseScore 4.

rvilarl:
@jaredjj3 @ronyeh I installed Dorico SE in the computer of my partner and this is how it looks :)

image

PR #164, opened 2023-11-07
Extract MusicXML examples from the W3C site's MNX specification

eliot-akira:
This pull request includes a script (based on @infojunkie's) to extract MusicXML examples from [W3C: MNX Specification - Comparing MNX and MusicXML](https://w3c.github.io/mnx/docs/comparisons/musicxml/).

The example files are in `tests/integration/__data__/w3c-mnx`.

I also wrote the logic for fetching each example's corresponding MNX (JSON) file and PNG image, but I made it optional in the script in case you only wanted the MusicXML files. Do you think these other formats would be useful to include?


jaredjj3:
@eliot-akira, thanks for the PR! This will help us get better coverage and not be completely indexed on the lilypond test suite.

>I also wrote the logic for fetching each example's corresponding MNX (JSON) file and PNG image, but I made it optional in the script in case you only wanted the MusicXML files. Do you think these other formats would be useful to include?

MNX yes, PNG probably not. Feel free to keep it there in case someone wants it to prep their workspace to develop offline or if we eventually find a use for it.

LGTM

eliot-akira:
Sure, I'll include the MNX files in the next commit.

When I tried it first, I downloaded them with the file extension `.mnx.json` to clarify what they are. Would you prefer just `.json`?

(Ah, I see the PR failed the lint step, I'll correct that also.)

jaredjj3:
Let's be more specific by choosing MNX to account for a possible future divergence from JSON.

eliot-akira:
Got it. I've updated the download script, and the files have been renamed with `.mnx.json`. I also ran prettier on them to pass the format check.

eliot-akira:
Oh, maybe you meant the file extension `.mnx` only?

From what I see in the discussion [MNX in JSON example documents and schema](https://github.com/w3c/mnx/discussions/292), it seems the format is meant to be strictly conformant to JSON, since they plan to use JSON schema with it.

jaredjj3:
>Oh, maybe you meant the file extension .mnx only?

Sorry, that's what I meant, but I guess using `.mnx.json`, we get the added benefit of prettier making it easier to read. Thanks for finding that discussion, looks like JSON is here to stay.

eliot-akira:
The `.json` extension helps with syntax highlighting too, in an editor or on the web. I like how the MNX examples are human readable (and writable, I suppose) - maybe even more so than MusicXML.

I finished reading through the MNX JSON discussion, and I understand it's a new idea with an unstable interface. (For example, I noticed they're using an object for `parts`, with the part name as the object property; but it probably should be an array like in MusicXML.) So the syntax will likely keep changing, and the examples are only useful for future reference.

Anyway, it's interesting to learn about. I think using JSON makes sense for easy parsing and portability.

---

Here's an idea from the discussion about rendering MNX JSON.

> For a real validation, my suggestion is: write some kind of toy renderer or even (dare I say it) toy editor for a subset of the proposed JSON version of MNX. No under the hood transformations to MusicXML allowed.

This relates to a question I had about `vexml`, whether its renderer can be used independently of MusicXML and XML elements. I imagine it's theoretically possible to implement a MusicXml-compatible document that uses JSON internally (or to be precise, JSON-serializable JS objects and arrays).

jaredjj3:
I'm on mobile so I can't provide direct links.

>This relates to a question I had about vexml, whether its renderer can be used independently of MusicXML and XML elements. I imagine it's theoretically possible to implement a MusicXML-compatible document that uses JSON-serializable objects and arrays internally, that can be passed to the vexml renderer.

The short answer is yes. Right now, the `rendering` is tightly coupled to `musicxml` (the internal sublibrary, not the MusicXML spec). If we created interfaces for the classes in that sublibrary, then the problem becomes "How do we translate MNX to MusicXML?" We wouldn't have to touch `rendering`.

However, we wouldn't be able to enjoy the benefits that mnx would bring, so maybe the conformance should be in the other direction: "How do we translate MNX to MusicXML?" From a strategic POV, this seems like the better choice since I imagine MNX would get developed more than MusicXML, but we'd have to wait for MNX to stabilize first.

Am I overlooking something important? What are your thoughts?

eliot-akira:
I'd guess it will take a while until the MNX specs stabilize, and even longer for it to start getting used by the public (applications and users). So it's probably too much of a moving target to plan around.

More in general, I was curious whether the internal representation of a MusicXML document (and all its components) in `vexml` could be JSON de/serializable.

If I remember correctly, OpenSheetMusicDisplay's internals are quite tied to working with XML elements, and it made it difficult to generate dynamic scores from application state. My solution at the time was to generate a MusicXML document in text, and let OSMD parse and render it every time the score changed. I suppose a more efficient way would be to generate OSMD classes directly - but it seemed to be a daunting task to map from JSON <-> OSMD classes.

> If we created interfaces for the classes in that sublibrary, then the problem becomes "How do we translate MNX to musicxml interfaces?" We wouldn't have to touch rendering.

Right, I see.. With this separation of `musicxml` and `rendering` modules, there's a possiblity to implement an `mnx` module with a compatible interface that can be passed to `rendering`.

For the purpose of dynamically generating scores, for example with a score editor or programmatically created examples, it seems useful to be able to work with JSON (plain JS objects, arrays) to go from JSON <-> vexml classes, without having to go through MusicXML in text form, nor XML elements which I imagine is a bit heavier to create than a plain object.

I've only started exploring vexml this week (but I've been watching its development along with vexflow 5). I found a nice collection of methods in `util/xml.ts` for dynamically creating a score. I'm trying it like:

```ts
import * as v from 'vexml/util/xml'

v.musicXml(
  v.scorePartwise({
    partList: v.partList({ scoreParts: [...] }),
    parts: [
      v.part({ id: 'P1', measures: [...] })
    ]
  })
)
```

That seems like it would be a fairly straight-forward mapping to JSON (and back?). Not necessarily related to MNX, but as its own serializable format.

sschmidTU:
@eliot-akira Just a quick note on OSMD: Yes, it's not easy to create the score model manually and yes, OSMD is basically made for reading MusicXML files instead of doing that, but it's possible.
We have an example in our sight reading sheet music generator that does this:
https://github.com/opensheetmusicdisplay/osme/blob/develop/src/OSME/SourceGenerator/ExampleSourceGenerator.ts

eliot-akira:
Thank you for the input, @sschmidTU, I love the work you're doing with OSMD.

Maybe my use case is out of scope for OSMD or vexml, because it's about generating scores with no actual MusicXML document. It might be better solved as a higher level of abstraction. For example, I have a database of scales from which static pages are generated (in Next.js), a page for each scale. On the page, it shows every "mode" of the scale, with notations starting from each note of the scale.

The easiest way, without having to figure out how to manually build up the score instance, was to generate MusicXML in text format. I used a little helper library called [xastscript](https://github.com/syntax-tree/xastscript), somewhat similar to how JSX works. But it seemed inefficient to go from a musical structure in JS (i.e., a scale as an array of intervals), to render it to XML, then pass it to OSMD (or vexml) to parse back into a structure in JS.

I guess MNX may eventually become useful for such a purpose, so that a score can be created and manipulated directly as plain objects and arrays. On the side of the renderer, I imagine this may not be simple to support - unless it was designed from early on to keep the renderer separate, without from directly interacting with XML elements. Maybe OSMD and vexml are already designed that way, to leave room for using a different underlying "element" (not a DOM element, but a virtual one) as long as it has a compatible interface.

Not sure how practical it would be, but just the idea of MNX seems like a good thing to consider for `vexml`, being a more newly designed library, there could be an opportunity for making dynamic score generation easier somehow.

jaredjj3:
>More in general, I was curious whether the internal representation of a MusicXML document (and all its components) in vexml could be JSON de/serializable.

@eliot-akira, it sounds one of the things you want is an easier way to edit MusicXML. Have you checked out [@stringsync/musicxml](https://github.com/stringsync/musicxml)? Each of the [generated elements](https://github.com/stringsync/musicxml/blob/master/src/generated/elements.ts) expose `attributes` and `contents` properties as POJOs that are compatible with JSON. You could convert to and from JSON if you wanted to, but the library already abstracts serializing/deserializing MusicXML from you.

When I started on this MusicXML journey, I initially didn't like dealing with XML. The schema felt hard to read and even more difficult to program with. However, I realized that I could wield the DOM APIs to navigate through them, which has been extremely useful.

Take a look at [NamedElement](https://github.com/stringsync/vexml/blob/3ccb1228cbafb45f7086bfd1e504425ff96af817/src/util/namedelement.ts). Implementing something like `ancestor` would not be as elegant dealing with JSON. Your cursor would have to store the parent of each node, adding complexity to you. Admittedly, it's not a lot of complexity.

>My solution at the time was to generate a MusicXML document in text, and let OSMD parse and render it every time the score changed.

I _think_ that when using dynamic measure widths, this is pretty much the best solution in terms of performance vs maintainability tradeoff.

The key problem is that all measure widths are dependent on each other. Let's say you update a measure and its width grows enough where you have to push a measure to another system. The measures in the same and next systems have to recalculate and resize BUT the measures in the preceding systems don't have to adjust.

You could come up with a smarter rendering model that only re-renders what's needed. But I feel like the maintenance burden would not be worth the performance cost. I haven't thought deeply about this, so I could be totally wrong about that.

It feels very similar to how browser layout engines deal with CSS, but I can't fully articulate the analogy. Maybe a better one is how React tries to only re-render components that changed by diffing its internal representation of the DOM.

jaredjj3:
@eliot-akira, thanks again for the contribution! If you ever want to work on MNX support in this library, please let me know. We will certainly build upon your work here.

eliot-akira:
Glad I was able to contribute. Looking forward to following the evolution of Vexflow and VexML. Appreciated the thoughtful reply on my tangential question about MNX and JSON representation.

> Each of the [generated elements](https://github.com/stringsync/musicxml/blob/master/src/generated/elements.ts) (heads up: big file) expose attributes and contents properties as **POJOs that are compatible with JSON**. You could convert to and from JSON if you wanted to, but the library already abstracts serializing/deserializing MusicXML from you.

Started reading through [@stringsync/musicxml](https://github.com/stringsync/musicxml), and the generated elements are amazing, what a massive library of typed MusicXML elements and attributes. This definitely covers everything I wanted and more for dynamically generating scores as MusicXML documents.

PR #169, opened 2023-11-18
Flesh out vexml.dev

### LINKED ISSUE #167: Create vexml site
## Requirements

**Make a simple site that allows users to render MusicXML documents. using `vexml`.**

- [x] It should be hosted on Github Pages.
- [x] It should provide a default rendered MusicXML document.
- [x] It should display rendering stats.
- [x] When the width of the `vexml` container changes, the score should re-render and update the rendering stats.
- [x] It may provide a MusicXML editor where changes are debounced. Errors are displayed nicely.
- [x] For now, it should the `master` branch as the `vexml` version. Maybe later, users can choose between `master` and releases.
- [ ] It has a link that allows users to easily file issues against. As a stretch goal, users should be able to easily export the rendered image, the MusicXML, and vexml version to the issue. Otherwise, make sure that the issue template requires them to include it.

## Hints

- This could probably be done in plain JavaScript. Try to avoid a whole application rendering pipeline for this.
- Nest everything under a `site` directory.
- WRT rendering stats, just do rendering time for now. Later, consider adding an object viewer that lets the user explore the resulting rendering tree.
- You might need to address #136 if you can't figure out how to compile the library with Github Pages' deployment pipeline. Then host a minified version on some CDN.
- Consider a simple component library like [Bootstrap](https://getbootstrap.com/) or [Tailwind ](https://tailwindcss.com/).

jaredjj3:
This PR sets up the rendering demo but does not have functionality for the "Report Issue" button. That will be implemented in a downstream PR. I want to merge this one so people can start experimenting with `vexml`.

image


jaredjj3:
I'll be providing snapshots as I progress.

image

rvilarl:
@jaredjj3 compressed musicxml does not seem to be supported and most of the files are compressed

jaredjj3:
Any idea how to handle this? My first instinct is to decompress it, then handle it as normal.

rvilarl:
Yes simply unzip and use the xml file in the root directory.

PR #176, opened 2023-11-27
Support multi-system spanners

jaredjj3:
This PR adds support for spanners that span multiple systems. This is not necessarily encoded into the document since it doesn't know anything about the viewport width.

## Tasks

- [x] Create `vexml` test suite and test data.
- [x] Pipe `Address` to spanner _fragments_ that can validly span multiple systems (wedges, pedals, vibratos, and octave shifts).
- [x] Update `Spanners` (and maybe `.*isAllowed` methods) to evaluate fragment addresses when evaluating what `vexflow` objects to create. When fabricating an "extra" spanner due to a system change, it needs to inherit the fundamental properties of the spanner and close it properly.

## vexml

image


jaredjj3:
@rvilarl, I potentially came across a `vexflow` bug developing this, but there's a good chance that this wasn't intended to be supported.

Two `vexflow.StaveHairpin` objects where the `firstNote` and `lastNote` are different `StaveNote` objects. This is working as expected.

image

**Five** `vexflow.StaveHairpin` objects where:

- the first four `firstNote` and `lastNote` are different `StaveNote` objects
- the last `firstNote` and `lastNote` are the _same_ `StaveNote` object

I expected a `vexflow.StaveHairpin` that spans the single note to be rendered, but I got a vertical line instead.

image

If this isn't intended in `vexflow`, WDYT of `vexflow` throwing an error instead of rendering a broken `vexflow.StaveHairpin`?

rvilarl:
I will look into the issue

PR #178, opened 2023-12-03
Add multiple examples to the site

jaredjj3:
This PR adds multiple examples to [vexml.dev](http://vexml.dev).

- [x] Use different file input on small viewports.
- [x] Make React code more idiomatic by removing unnecessary `useCallback`.
- [x] Add `<Select>` component.
- [x] Reads `*.musicxml` files from `src/examples` dir and wires it into the `<Select>` component.


jaredjj3:
@rvilarl, just a heads up. You can add `*.musicxml` files to `site/src/examples/**`. If you have some good examples that you think ought to be there, feel free to add.

PR #179, opened 2023-12-03
Render lilypond41a

### LINKED ISSUE #173: [BUG] (v0.0.0): staves in scores with more than one instrument are overlapping
**Expected behavior**
image

**Actual behavior**
The piano and basson staves overlap. No error message.

**Screenshot**
image

**Hints**
The more you provide, the faster we can fix it.

- [X] Attach the MusicXML file used to produce the bug.
- [ ] Attach the error messages.
- [ ] Include any `vexml` code references where the bug may be.

[Sarabande_de_Haendel.zip](https://github.com/stringsync/vexml/files/13403920/Sarabande_de_Haendel.zip)

jaredjj3:
This PR fixes #173.

- [x] Render `lilypond41a`.
- [x] Render part names.
- [x] Add stave connectors from the first part to the last part.
- [x] Account for brace stave connectors when calculating `x, y` for rendering.
- [x] Render [Sarabande_de_Haendel.zip](https://github.com/stringsync/vexml/files/13403920/Sarabande_de_Haendel.zip).

## vexml

41a-MultiParts-Partorder_900px


jaredjj3:
Some progress:

Before:

vexml_dev_0_0_0 (1)

After:

vexml_dev_0_0_0

jaredjj3:
Sarabande is looking better, but it has a few more issues:

image

The last thing I'm fixing in this PR is the widths of the second part. It should line up with the first part, but for some reason it doesn't. When I manually change the `targetSystemWidth`, it proportionally changes and doesn't fill out the whole system for some reason.

rvilarl:
@jaredjj3 if you use a single system and add three staves it should work. I believe.

jaredjj3:
Ah thanks, however, it looks like I need to provide the `vexflow.System` constructor with a `vexflow.Factory` ([code](https://github.com/0xfe/vexflow/blob/7e7eb97bf1580a31171302b3bd8165f057b692ba/src/system.ts#L96C82-L96C82)). I avoided using `vexflow.Factory` because I wanted maximum flexibility on how things are rendered, so I went with the lowest level structures.

I'll study how [vexflow.System.format](https://github.com/0xfe/vexflow/blob/7e7eb97bf1580a31171302b3bd8165f057b692ba/src/system.ts#L238) works and try to apply it to `vexml`.

jaredjj3:
I figured it out. So there's this rendering variable called `minRequiredSystemWidth` and it's used to determine how much to stretch each measure in a given system. The problem was that I was calculating this one for each measure _across all parts_. However, since each measure doesn't know anything about the other measure's part, we can't compute the `minRequiredSystemWidth` calculation like it does.

image

I'll fix formatting in another PR, probably using a simpler score.

rvilarl:
Why don't you want to use factory?

jaredjj3:
This is a really difficult design decision to communicate. Let me know if you still need clarity.

The main issue is that `Factory` would force me to convert from MusicXML to `vexflow` objects in a ~specific order. The rendering strategy has been to convert to `vexflow` objects ASAP when I have enough context. Then, at the end, I can do the same thing factory does: set the context and draw everything.

Ultimately, I don't want `vexflow` tightly coupled to the implementation of `vexflow.Factory`. I think that API makes more sense when the developer already knows all the structures that should be written. MusicXML doesn't organize this information in a convenient manner (which is inherent to the complex nature of music), so I don't have the benefit of "knowing" exactly what `vexflow` objects are needed until I go through the entire MusicXML document — that's why I'm creating `vexflow` objects ASAP.

jaredjj3:
@rvilarl, if you're familiar with AWS, I think using this is similar to using ECS, Elastic Beanstalk, RDS, ELB, etc. (Factory) instead of EC2 (vexflow objects). The higher abstractions (ECS, Elastic Beanstalk, RDS, ELB, etc.) are easier to use, but they're not as flexible if you managed the low level EC2 instances yourself. I want the maximum flexibility that `vexflow` can offer.

PR #198, opened 2024-01-05
Support complex formatting

### LINKED ISSUE #189: [BUG] (v0.0.0): <Sarabande measure 8>
**Expected behavior**
image

**Actual behavior**
Measure 8:
- on the first staff the accidental does not need to be repeated.
- on the second staff the voices are not formatted correctly.
- on the third staff there are too many rests (they might be hidden in the score)

**Screenshot**
image

**Hints**
Sarabande alreay added on previous issues.

- [ ] Attach the MusicXML file used to produce the bug.
- [ ] Attach the error messages.
- [ ] Include any `vexml` code references where the bug may be.

jaredjj3:
This PR fixes #189 by fixing how divisions are tracked when creating voices.

## vexml

image

## osmd

image

## MuseScore 4

image

## Guitar Pro 7

image


jaredjj3:
For the second and third staves, if I render only the first or only the second voice of each stave, the formatting is correct. There must be a weird interaction between the voices that I'm not accounting for.

_first voice only for staves 2 and 3_

image

_second voice only for staves 2 and 3_

image

jaredjj3:
`vexflow.Voice.setStrict(true)` hints at the problem.

```
16:59:16.416 (1116px) Error: [RuntimeError] IncompleteVoice: Voice does not have enough notes.
```
 
This is what @sschmidTU was talking about in https://github.com/stringsync/vexml/pull/162#issuecomment-1791103875. Right now, I'm ignoring notes with `[print-object="no"]`, but this is yielding the wrong result.

Instead of completely ignoring the notes with `[print-object="no"]`, I'm going to update the voice generation algorithm to use `rendering.GhostNote` instead, then post the result.

jaredjj3:
I still get the incorrect result and `vexflow` also complains if I set the voice strict mode to `true`. However, I think I'm moving in the right direction.

image

Ultimately, I'm going to revisit the voice generation algorithm and I'm going to make `vexml` try its best to fill gaps with `GhostNote` objects. If there are too many notes, it's infeasible to try to pick which notes to cull out. That's ok, because I want to be able to support measures with invalid beats to be editor friendly, where invalid measures would be a common occurrence.

jaredjj3:
Looking at the beats for each voice's entries helps illuminate the problem:

```
voiceId 1: (start: 0, end: 2),(start: 2, end: 4),(start: 4, end: 5),(start: 5, end: 6)
voiceId 3: (start: 2, end: 3),(start: 3, end: 4),(start: 4, end: 5),(start: 5, end: 6)
voiceId 5: (start: -6, end: -4),(start: -4, end: -2),(start: -2, end: -1),(start: -1, end: 0)
voiceId 7: (start: -4, end: -3),(start: -3, end: -2),(start: -2, end: -1),(start: -1, end: 0)
```

Voices 1 and 3 seem correct. Voices 5 and 7 are definitely invalid — it looks like they're offset by 6 beats. It might be the way `vexml` is handling `<backup>` elements...

jaredjj3:
Looks like the rests are rendering in the correct place according to `<display-step>` and `<display-octave>`.

rvilarl:
> Looks like the rests are rendering in the correct place according to `<display-step>` and `<display-octave>`.

@jaredjj3 have you tried `alignRests: true` as option to `format`?

jaredjj3:
I just tried that formatting option, but I get the same result. Any other hints?

rvilarl:
I will have a look.

sschmidTU:
> This is what @sschmidTU was talking about in [#162 (comment)](https://github.com/stringsync/vexml/pull/162#issuecomment-1791103875). Right now, I'm ignoring notes with `[print-object="no"]`, but this is yielding the wrong result.
> 
> Instead of completely ignoring the notes with `[print-object="no"]`, I'm going to update the voice generation algorithm to use `rendering.GhostNote` instead, then post the result.

Didn't you want to just render the notes with invisible color, like I recommended? Then they can be made visible via SVG (setting the color), and there will also be no issues with using GhostNotes. Just a question out of curiosity.

jaredjj3:
I experimented with transparent StaveNotes as you suggested first, but then I realized that they are functionally acting the same as the GhostNotes that come from different voices when applying `<backup>` and `<forward>` elements. It also seemed that GhostNote objects only need a subset of StaveNote data, so I wanted to keep things slim until something else warrants otherwise. I may end up doing what you originally suggested later on.

rvilarl:
I missed that.

 '`note.renderOptions.draw = false` should make any note invisible.

rvilarl:
@jaredjj3 did you see my comment above?

jaredjj3:
Ah, now I have. Does `GhostNote` offer any advantage over a `StaveNote` using `note.renderOptions.draw = false`?

rvilarl:
I do not know. @ronyeh can you help?

PR #204, opened 2024-01-19
Render lilypond32a (accents)

jaredjj3:
This PR renders accents, which is part of lilypond32a.


jaredjj3:
I decided to take a different approach with this. Instead of making a `musicxml` class for each of the `<articulations>` [children](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/articulations/), I'll just make one for `<articulations>` itself and make a `rendering.Articulation` class.

PR #211, opened 2024-02-17
Honor measure widths

### LINKED ISSUE #210: Honor measure widths
**Expected behavior**
Render measure widths when explicitly specified.

image

**Actual behavior**
vexml ignores the widths and calculates its own.

**Screenshot**
A screenshot of the `vexml` rendering.

image

**Hints**
[Prelude_No._1_BWV_846_in_C_Major.zip](https://github.com/stringsync/vexml/files/14296970/Prelude_No._1_BWV_846_in_C_Major.zip)

It looks like the widths are based on the measure fragment widths in [Seed.split](https://github.com/stringsync/vexml/blob/c477af4ba64f2d02d2fcff35adf58acff2e51df5/src/rendering/seed.ts#L37).

- [x] Attach the MusicXML file used to produce the bug.
- [x] Attach the error messages.
- [x] Include any `vexml` code references where the bug may be.

jaredjj3:
This PR fixes #210 by honoring measure widths when they are specified. When a measure width is "honored", it means its using the width as the _basis_ of calculation. When a measure width is not specified, `vexml` will continue to calculate the width as normal.

- Fixes a bug where measures with multiple staves would be offset incorrectly.

## vexml

vexml_dev_0_0_0


rvilarl:
@jaredjj3 Some measures seem not to fit 10, 12, 28. There are overlaps from notes on different staves. I thimk that MusicXML also specifies this distance, is that right?

jaredjj3:
@rvilarl, yeah and the overlapping gets worse the smaller the viewport. MusicXML does define [`<distance>`](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/distance/), but this isn't specified in [Prelude_No._1_BWV_846_in_C_Major.zip](https://github.com/stringsync/vexml/files/14296970/Prelude_No._1_BWV_846_in_C_Major.zip). There is a [`<scaling>`](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/scaling/) element specified, so I wonder if that will correct the sizing issues. Before I invest more time into this, do you see any other things I could do to fix the sizing?

rvilarl:
I will have a look.

jaredjj3:
@rvilarl , after thinking about it, I think my approach in this PR was wrong. In [3fb4d04](https://github.com/stringsync/vexml/pull/211/commits/3fb4d045e7f1b0d8d828dbe648f1e723d8f18c92), I updated the system-generating process to use the measure's specified width as the basis for how much width it _requires_ to render. In reality, the required width isn't changing — the _target_ width is. I'm telling the procedure "trust me, I know the real width it requires".

When `vexml` resizes the widths to fit in the system, we end up with measures that don't have enough width to accommodate all the notes.

When I finish #212, I plan to revisit this:

- Update the system generating procedure to use the measure specified width as a target width instead of a required width.
- Account for `<scaling>` when calculating how many pixels a `<measure>` element's width is.

After you look at it, let me know if you have other thoughts.

PR #212, opened 2024-02-18
Render lilypond31a

jaredjj3:
This PR renders lilypond31a.

- Unsupported elements: `<eyeglasses>`, `<other-dynamics>`, `<dashes>`, `<bracket>`, `<harp-pedals>`, `<damp>`, `<damp-all>`, `<scordatura>`, `<accordion-registration>`
- `<rehearsal>` elements render with plain font and ignore the `enclosure` attribute.
- Coda symbols render at the beginning of the measure.
- Dynamics symbols render poorly when close together.

## vexml

vexml_dev_0_0_0 (2)


jaredjj3:
Going forward, I'm going to leave a better paper trail on decisions I make.

I ran into an issue rendering [`<rehearsal>`](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/rehearsal/) elements.

## lilypond

image

## OSMD

image

## MuseScore 4

image

`<rehearsal>` elements have an `enclosure` property, which determines what shape encloses the rehearsal name.

>The enclosure is square if not specified.

The [Box Text](https://github.com/0xfe/vexflow/blob/8ddc8fa1a6d304a879e73830919fa17f3a9bdef4/tests/textformatter_tests.ts#L93) vexflow test demonstrates how to use a `vexflow.TextFormatter` to determine the dimensions for rendering text and their enclosures. However, this is more low level than what I anticipated.

In order to properly support this, I think I should extend `drawables.Text` to be able to render enclosures. For now, I'm going to figure out the best way to quickly render the `<rehearsal>` value above the note it gets attached to without worrying about the enclosure.

jaredjj3:
I ended up going with `vexflow.Annotation`. If I see the need to render something around an annotation, I should consider making a request from `vexflow` to support it instead of rolling it out my own.

## vexml

image

jaredjj3:
[`<direction-type>`](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/direction-type/) allows multiple `<segno>` elements. I'm not sure I understand what it means to have multiple segnos. I decided to just render a single segno per measure-part.

## vexml

image

jaredjj3:
I made the same decision for codas.

## vexml

image

jaredjj3:
I ran into an issue rendering dynamics. From what I can tell, they're rendered as [vexflow.TextDynamics](https://github.com/0xfe/vexflow/blob/8ddc8fa1a6d304a879e73830919fa17f3a9bdef4/src/textdynamics.ts#L23), which is a type of `vexflow.Note`. The problem with rendering is that I cannot simply attach a dynamic to a note like a `vexflow.NoteModifier`. It seems like I need to create a new parallel voice of  notes to house dynamics.

@rvilarl, what's a recommended way to associate a dynamic with a note in `vexflow`?

jaredjj3:
I found a `vexflow` [test](https://github.com/0xfe/vexflow/blob/8ddc8fa1a6d304a879e73830919fa17f3a9bdef4/tests/textnote_tests.ts#L155) that demonstrates creating another voice for `vexflow.TextNote` objects that appear directly over `vexflow.StaveNote` objects. It's the caller's responsibility for calculating the durations accordingly.

I think I should use [rendering.VoiceEntryData](https://github.com/stringsync/vexml/blob/b5144be727d497b59ed374bc6c5a0f7732ce9cc4/src/rendering/chorus.ts#L29) to figure out how many times I need to clone a voice, and use `vexflow.GhostNote` objects to serve as placeholders.

I'm thinking of introducing a new data structure to house non-`vexflow.StaveNote` objects.

inputs
- `<direction>` elements and their corresponding start `Division`
- `vexflow.GhostNote[]` mirroring the note durations of the `StaveNote` voice

outputs
- `vexflow.Voice` with the `vexflow.TextNote` and other `vexflow.Note` subclasses that correspond to directions, attached to the notes correctly.

I think I'll start by making `rendering.Voice` aware of the directions it contains. Then, I'll implement a way to create a placeholder clone, inserting `vexflow.TextNote` objects where they belong.

jaredjj3:
When fleshing out some of the details of the placeholder `rendering.Voice`, I realized that `rendering.Chorus` is doing too much. `rendering.Voice` should produce multiple `vexflow.Voice` objects, and should ultimately be in charge of associating `musicxml.Direction` elements with `rendering.VoiceEntry` or the placeholder `rendering.Voice`.

**Action Items**

- Update `rendering.Voice` to be constructed with `rendering.VoiceEntry` data.
- Update `rendering.Voice` to house the `rendering.GhostNote` computation.
- Update `rendering.Voice.render` to create placeholder `vexflow.Voice` objects as needed to fulfill `musicxml.Direction` elements that must be rendered in a `vexflow.Voice`.
- Update `rendering.Voice.render` to handle rendering the new `rendering.VoiceEntry` elements.

jaredjj3:
I'm almost finished fixing the `Chorus` and `Voice` structures. The last things that are broken are:

- Octave shifts are not affecting where notes are placed. This is expected, because I wanted to move the concept of octave shifts to `StaveSignature` since it affects note placement. Before this PR, octave shifts were incorrectly calculated at the `MeasureFragment` level. This would yield the wrong outcome if an octave shift spanned more than a single `MeasureFrgament`.
- Spanners are not being rendering correctly. I suspect that `<direction>` elements are not getting associated correctly, but I currently don't know exactly what's wrong.
- I've added TODOs to the new `Chorus` class. I'll address those before moving forward with this PR.

Over the next week, I'll have a few days where I can focus on `vexml` completely. I'm going to try to get the `rendering` library done and then start working on the cursor.

jaredjj3:
https://github.com/stringsync/vexml/pull/212#issuecomment-1962447735 more-or-less worked:

image

There are some formatting issues, but that can be addressed later. It's probably unlikely that dynamics would change that much in music.

Issue #78, opened 2023-01-17
Challenge: Generate code for Joplin - Gladiolus Rag


rvilarl:
The good news: no crash reading the XML file.
The challenge: quite a few things to fix.

![image](https://user-images.githubusercontent.com/22865285/212888556-bdd2c253-5362-4244-8086-69c75041e59f.png)

rvilarl:
Our target :)

![image](https://user-images.githubusercontent.com/22865285/212888811-569173d4-ecc4-4e8d-971f-cef1808cf05c.png)

jaredjj3:
I like this approach. It gives us concrete bugs to work on. The cross-staff notations are particularly interesting.

jaredjj3:
@rvilarl, would you attach the MusicXML file somewhere in this issue please?

rvilarl:
The score is this one https://musescore.com/user/58480/scores/1749906

rvilarl:
@jaredjj3 the score produces this fault at https://vexml.dev/
 
[Gladiolus_Rag__Scott_Joplin__1907.zip](https://github.com/stringsync/vexml/files/13811849/Gladiolus_Rag__Scott_Joplin__1907.zip)


17:12:14.620 (1296px) Error: [RuntimeError] BadArguments: Too few notes for beam. at new sf (https://vexml.dev/assets/index-e7Ei05G8.js:60:246688) at Wo.render (https://vexml.dev/assets/index-e7Ei05G8.js:60:1213988) at https://vexml.dev/assets/index-e7Ei05G8.js:60:1224036 at Array.map (<anonymous>) at Jh.render (https://vexml.dev/assets/index-e7Ei05G8.js:60:1224027) at Zs.render (https://vexml.dev/assets/index-e7Ei05G8.js:61:5105) at mr.render (https://vexml.dev/assets/index-e7Ei05G8.js:63:174) at https://vexml.dev/assets/index-e7Ei05G8.js:63:7625 at Dc (https://vexml.dev/assets/index-e7Ei05G8.js:40:24283) at Ss (https://vexml.dev/assets/index-e7Ei05G8.js:40:42406)

jaredjj3:
Good catch, I think we can fix this by adding a check [here](https://github.com/stringsync/vexml/blob/cff66f352d84ecd8102228163716250a0330fa39/src/rendering/beam.ts#L84) and making the `BeamRendering.vexflow.beam` nullable.

rvilarl:
[...]

PR #223, opened 2024-05-28
Gladiolus Rag fixes

jaredjj3:
This PR addresses some issues rendering Gladiolus Rag by Joplin in #78, but does not completely fix it.

- [x] Ensure that the top of a rendering is present.
- [x] Prevent ties from rendering across systems.
- [x] Investigate increasing `StaveTie` height, part distance, or both.
- [x] Render repeats.

This PR adds basic support for rendering repeat barlines. I'll address alternative endings in lilypond 45b-RepeatWithAlternatives.xml.

## Before

vexml_dev_0_0_0 (1)

## After

vexml_dev_0_0_0


rvilarl:
wow! This looks so good!

rvilarl:
The cross tab elements are missing.

image

jaredjj3:
Thanks for calling this out — this part is very tricky. The next step is to make a small [vexml integration test case](https://github.com/stringsync/vexml/tree/master/tests/integration/__data__/vexml) reproducing that.

I have a theory on what's broken.

Beams assume that they will see all their notes before starting another beam. I know this because they are not [keyed](https://github.com/stringsync/vexml/blob/f7b839a0596cef76a106df48e55274d86cfae35a/src/rendering/beam.ts#L26) by anything in their underlying `SpannerMap`. There's currently no mechanism to have multiple beams open at the same time.

So far, I can think of doing one of the following solutions:

1. Figure out a key for the beam's `SpannerMap`. Maybe we can infer one from the note's [Address](https://github.com/stringsync/vexml/blob/master/src/rendering/address.ts). The thing is that I'd have to be very careful with further assumptions about beams (e.g. can beams traverse measure boundaries?).
2. When providing `SpannerData` ([code](https://github.com/stringsync/vexml/blob/f7b839a0596cef76a106df48e55274d86cfae35a/src/rendering/types.ts#L6)), include data from all staves. Then, `Beam` should be responsible for searching other staves for notes that also belong to a given beam opening.

rvilarl:
I believed not, but...

image

jaredjj3:
I'm glad you found this! Music notation is full of wonderful edge cases.

Issue #158, opened 2023-10-18
Create mechanism to programmatically navigate rendering


jaredjj3:
## Requirements

From #118, #122, and myself, the requirements are:

- The cursor can accept position instructions. Callers should be able to specify time, position (x, y), measure index/number with or without repeat information, next(n), previous(n), and the cursor should update its position accordingly.
- `vexml` can generate default position instructions coupled to time.
- The cursor provides information about the active elements, including provenance information about the MusicXML that is associated with the active elements.
- The cursor emits events: `activeelementchange`, `positioninstructionreceived`, `cursorexhausted`, etc..
- The cursor's visibility is toggleable.
- A rendering can accommodate multiple cursors. Each cursor is agnostic to one another and therefore it is the caller's responsibility to coordinate them.
- Cursors can be styled.
- Cursors can be interpolated between voice entries or they can snap to the closest voice entry given the last position instruction.
- Cursors can navigate through entire voice entries or they can navigate individual elements _within_ a voice. The former is typically more useful for reading music, while the latter is more useful for editing MusicXML.

## Hints

I would like something that is derived from the `*Rendering` object tree that accomplishes all of these things. Right now, the `*Rendering` object tree expose the underlying `vexflow` objects, but I would like callers to not have to interact with that.

I recommend creating a new library within this project called `cursor`, which will house all the logic needed to accomplish the requirements.  Consider converting the `*Rendering` object tree to something more useful for the cursor as a first step, which will also have the benefit of decoupling `cursor` from `rendering`. That way, if `rendering` has to change significantly, the `cursor` library should only need to be updated in one place.

## Inspiration

- [opensheetmusicdisplay/opensheetmusicdisplay](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay)
- [infojunkie/musicxml-player](https://github.com/infojunkie/musicxml-player)
- [stringsync/stringsync](https://github.com/stringsync/stringsync)

jaredjj3:
@infojunkie, I think it'll be `O(months)` before I really nail down rendering, so I invite you to start looking into this. What do you think is a good strategy for dealing with `*Rendering` object tree ([`ScoreRendering`](https://github.com/stringsync/vexml/blob/cbf8ec451f140e6d48585c580c64e3fb5259685b/src/rendering/score.ts#L14) as the root) changes that would break this work?

infojunkie:
Thanks for inviting me! This is a big subject, and it can get overwhelming quickly.

[...]

PR #227, opened 2024-06-03
Create event system

### LINKED ISSUE #159: Music sheet events
## Requirements

When the user interacts (click, tap, keyboard press during focus, etc.) with a rendered score, the score should emit events that can be hooked into. An event may or may not contain information about what was interacted with.

>**NOTE:** Any events tied to a cursor should be emitted by the cursor instead. See #158.

## Hints

Before developing anything, I highly recommend that you make a list of events + payloads (see MusicDisplay Event definitions below).

There is prior art in stringsync.

- [EventBus](https://github.com/stringsync/stringsync/blob/c65bc9d10f5392ac69b1700488393340ff0686b2/web/src/lib/EventBus/EventBus.ts): Pubsub mechanism for events with payloads.
- [MusicDisplay Event definitions](https://github.com/stringsync/stringsync/blob/c65bc9d10f5392ac69b1700488393340ff0686b2/web/src/lib/MusicDisplay/types.ts#L49): A list of all the different events and their payloads built on top of osmd.
- [SVGEventProxy](https://github.com/stringsync/stringsync/blob/c65bc9d10f5392ac69b1700488393340ff0686b2/web/src/lib/MusicDisplay/svg/SVGEventProxy.ts): An interaction sink that translates DOM events to MusicDisplay events.
- [State machine for the pointer](https://github.com/stringsync/stringsync/blob/c65bc9d10f5392ac69b1700488393340ff0686b2/web/src/lib/MusicDisplay/pointer/pointer.ts): The SVGEventProxy pipes DOM events to this, and this is what ultimately calculates the MusicDisplay events.
- [MusicDisplayLocator](https://github.com/stringsync/stringsync/blob/c65bc9d10f5392ac69b1700488393340ff0686b2/web/src/lib/MusicDisplay/locator/MusicDisplayLocator.ts): Efficiently finds a osmd cursor snapshot given an (x, y) position.

jaredjj3:
This PR fixes #159.

It introduces some new modules:

- `cursors`: This is a public-facing module that helps `vexml` and developers interact with the music sheet. Currently, there is only a `PointCursor`. Later, this will house `MusicalSequenceCursor` and `KeyboardNavigationCursor` (I haven't landed on naming yet).
- `spatial`: This module models different bounding box shapes and introduces a quad tree implementation for locating these shapes. Right now, `vexflow.StaveNote` objects are modeled as a single rectangle, but `vexml` supports composite modeling such as a circle + rectangle.
- `events`: This module contains utilities that synthesizes events from native events. It does not contain the actual event declarations, because I didn't want it to know about the `rendering` module. Instead, I declared all the events in `rendering/events.ts`.


jaredjj3:
@alindsay55661, it's been a while since we talked on #158. I wanted to let you know that I'm wiring music sheet event support in this PR. Feel free to leave comments if you see an opportunity to improve the approach!

alindsay55661:
> @alindsay55661, it's been a while since we talked on #158. I wanted to let you know that I'm wiring music sheet event support in this PR. Feel free to leave comments if you see an opportunity to improve the approach!

I really like the strategy you are using (single listener) and the support you've provided via READMEs and tests. Based on a cursory review, it looks like you are planning to expose all the potential click collisions in the payload and the consumer gets to decide what to do with them? Or are you planning to refine the event further prior to dispatch (is this even possible in all cases)?

One thing that comes to mind is, depending on where the bounding boxes lie, how does a consumer distinguish between clicking on an accidental vs. a notehead? Or any other number of proximately close elements that may have overlapping rectangles, (perhaps even other notes, stems, flags, etc.?). 

The second question I have is around #158. Will the event system differ when triggered by playback cursors instead of mouse cursors. Is the idea to try and maintain some parity in the payloads of these differing experiences, or do you plan to embrace the fact that playback will practically dispatch only a subset of interaction events? For example, playback cursors will not likely emit events for the dots of a dotted note or the accidentals (independently) of a note, while interaction events will. Just thinking out loud in real-time here and wondering what I would expect as a consumer... I'm not sure.

Thank you for all your hard work on this project!

jaredjj3:
Awesome, thanks for the feedback so far, you've uncovered some things I haven't thought about.

>Based on a cursory review, it looks like you are planning to expose all the potential click collisions in the payload and the consumer gets to decide what to do with them? Or are you planning to refine the event further prior to dispatch (is this even possible in all cases)?

This is one of the things I'm worried about. Short term, I'm just going to pick the element whose bounding box's center has the smallest distance from the event.

However, I realize pitfalls of the short term approach. The distance from the centroid of an element may be inaccurate for something like a slur. I think a better approach would be to create anchor points that characterize the element.

image

As you can see, the anchor points are contained in the bounding box. I would then pick the element that has an anchor point with the smallest distance from the event to determine what element gets selected. I could still expose all the elements whose bounding box contains the event in some metadata, so the user can decide if they want to do anything with that or not.

>Will the event system differ when triggered by playback cursors instead of mouse cursors. Is the idea to try and maintain some parity in the payloads of these differing experiences, or do you plan to embrace the fact that playback will practically dispatch only a subset of interaction events?

Great point — there's an opportunity for me to reuse/extend the mouse cursor solution for playback cursors. I haven't thought about this deeply. Ultimately, I think I need to enumerate the events from each and determine the overlap. I'll do that early on so I can determine 

My gut feeling is that these are fundamentally too different to have significant overlap. The mouse cursor is essentially a point while the playback cursor is an area. I think the playback cursor will have to rely on the structure of the `rendering.*Rendering` objects to know what elements are in scope.

jaredjj3:
While developing this, I realized something particularly illuminating. A cursor is a dynamic scope that calculates what `vexflow.Elements` are within it.

- A cursor that moves _discretely_ (e.g. opensheetmusicdisplay [demo](https://opensheetmusicdisplay.org/demos/public-typescript-demo/)) is **driven by a trigger**. Its scope is defined by a sequence of `vexflow.Tickables` and the current index within that sequence.
- A cursor that moves _~continuously_ as a musical piece is played is **driven by time**. Its scope is defined by a sequence of `vexflow.Tickables` and the current time in the piece.
- A cursor that moves as a mouse or touch pointer moves is **driven by position**. Its scope is defined by a region that characterizes the pointer and whatever `vexflow.Element` regions intersect it.

I contemplated the usefulness expressing this as an abstraction:

```ts
interface CursorDriver {}
class SequenceCursorDriver implements CursorDriver {}
class TimeCursorDriver implements CursorDriver {}
class PositionCursorDriver implements CursorDriver {}

interface CursorScope {}
class SequenceCursorScope implements CursorScope {}
class PositionCursorScope implements CursorScope {}

class Cursor {
  constructor(
    private driver: CursorDriver,
    private scope: CursorScope,
  )
}
```

...but then I realized that there are some awkward combinations. For example, what does a `Cursor` with a `SequenceCursorDriver` and a `PositionCursorScope` even mean? The `SequenceCursorDriver` would not emit any positional data, so it seems like this is not the correct abstraction.

For now, I'm planning to just create three separate cursors. If any patterns arise from that, I'll carefully reconsider a common abstraction. This is probably better practice anyway.

```ts
class SequenceCursor {}
class TimeCursor {}
class PositionCursor {}
```

jaredjj3:
I have a very rough prototype of a working event system, but there's a lot that needs to be done.

## TODO

A stream of consciousness, in no particular order...

- [ ] Rethink the `spatial.Region` solution. Ultimately, we need a way to store areas in a quad tree. When I click a point, I want to be able to quickly query all the bounding boxes that contain it. Maybe I could use the quad tree as very coarse filtering mechanism then use `spatial.Region.contains` and `spatial.Region.getNearestPoint` as finer filtering mechanisms.
- [ ] Assuming the `spatial.Region` solution is kept, update some `rendering.*Rendering` objects to calculate `spatial.Region` objects. Maybe just note renderings for now. Maybe this warrants full classes, so we don't have to pollute rendering logic with computations needed for events? But then, where would the classes live?
- [ ] Determine a better home for `spatial.QuadTree` building.`vexml.ts` should be short and easy to read since it's the entrypoint of the library. Maybe `rendering.ScoreRendering` could build the quad tree.
- [ ] Update `spatial.QuadTree` construction to use bounds from the `rendering.ScoreRendering`, not the DOM `Element` container (since it could have styling that affects client rect like padding).
- [ ] Update `spatial.Rectangle` naming to `spatial.Rect`.
- [ ] Create convenience method for creating a `spatial.Rect` from a `vexflow.Box`.
- [ ] Consider adding a `spatial.Rect.corners` method.
- [ ] Consider creating a mechanism to support different bounding box shapes. Maybe some common interface like `Shape`. Something should know how to take two `Shape` objects and determine if there's a collision.
- [ ] Add proper types to `rendering.Rendering.cursor`. You might need to move all the `rendering.*Rendering` types to `rendering/types.ts` if the cursor is intended to return `rendering.*Rendering` objects.
- [ ] Map the dependency of `vexml` events to native events. You should only listen to a native event when a `vexml` event needs it.
- [ ] In `cursors.PointCursor`, lazily query the quad tree — don't do it on `cursors.PointCursor.update`.
- [ ] In `cursors.PointCursor`, formalize the "interaction area" concept and consider making it configurable.
- [ ] Fill out the intent document for the `cursors` module.
- [ ] Make the `'mousemove' | 'touchmove'` event throttling configurable per rendering through `rendering.RenderOptions`.
- [ ] Expose a mechanism to uninstall all events without needing to track all the handles.
- [ ] Consider capturing `rect.left` and `rect.top` `rendering.Rendering.getPoint` in member variables so they don't have to be recalculated every time the mouse moves.
- [ ] Lazily calculate `util/device`, maybe exposing it in a singleton class. From a unit testing standpoint, I'd have to mock a module, which isn't ideal. I'd rather be able to inject whatever `Device` I want independently of the environment.
- [ ] Get rid of unused cruft from `util/device` — we should only use what we need.

https://github.com/stringsync/vexml/assets/19232300/21098855-815f-45b1-bb7d-14566836dca2

jaredjj3:
The system works!

You can compose a list of `spatial.Shape` objects into an `rendering.InteractionModel`, which describe the boundaries of where a rendered object can be interacted with. Right now, I only support modeling a `rendering.StaveNoteRendering` with a `spatial.Rect` and `spatial.Circle`. 

> [!NOTE]  
> Let `n` be the number of shape types. The implementation size of `spatial.CollisionSubject` will be O(n<sup>2</sup>). I don't plan on adding other shapes right now.

I just wanted this PR to be scoped to getting an MVP, so I won't add a comprehensive list of events/features now. I'll do that in downstream PRs. However, I do plan to finish the remaining things in https://github.com/stringsync/vexml/pull/227#issuecomment-2156801946 before merging this.

https://github.com/stringsync/vexml/assets/19232300/42531016-88f5-40c0-aac7-54b30d0273a9

PR #230, opened 2024-06-16
Add config knobs and basic events

jaredjj3:
This PR expresses configuration in a way that allows callers to programmatically generate knobs for it. It also adds basic vexml events that can be listened to.

Note hit areas are now modeled as "lollipops" — a rect for the stem and a circle for the notehead.

image


jaredjj3:
Quick update: mouse enter and exit are working! I will eventually need help auditing the `rendering/events.ts` file to make sure that I'm exposing the right data and computing things correctly.

https://github.com/stringsync/vexml/assets/19232300/7672c7dd-9201-4d11-868e-e646da9ee506

jaredjj3:
I'm quickly seeing the limitations of canvas. I'm changing the `vexflow.StaveNote` colors like this:

```ts
const onEnter: vexml.EnterEventListener = (event) => {
  container.style.cursor = 'pointer';

  const value = event.target.value;
  if (value.type === 'stavenote') {
    value.vexflow.staveNote.getSVGElement()?.remove();
    value.vexflow.staveNote.setLedgerLineStyle({ strokeStyle: STRINGSYNC_RED });
    value.vexflow.staveNote.setStyle({ fillStyle: STRINGSYNC_RED, strokeStyle: STRINGSYNC_RED }).draw();
  }
};
const onExit: vexml.ExitEventListener = (event) => {
  container.style.cursor = 'default';

  const value = event.target.value;
  if (value.type === 'stavenote') {
    value.vexflow.staveNote.getSVGElement()?.remove();
    value.vexflow.staveNote.setLedgerLineStyle({ strokeStyle: 'black' });
    value.vexflow.staveNote.setStyle({ fillStyle: 'black' }).draw();
  }
};
```

While inefficient, SVG has a clean replacement when re-drawing the note. However, canvas has to render a note on top of the previous one, and there are black artifacts around the note. While this is expected, I thought it would look a little better than this.


https://github.com/stringsync/vexml/assets/19232300/0cafe331-ce8b-46aa-b632-667ca14c8308

rvilarl:
Hi Jared, it seems that the lollipops are bigger than the vexml rectangles, what is the advantage?

jaredjj3:
That's a fair critique. Why overrepresent the shape if the box is simpler and adequately contains the symbol? My answer is that I think the lollipop is a higher resolution bounding shape of a stemmed note.

Given a point on the rendered music sheet, there may be several bounding shapes that contain the point. I needed a way to sort the shapes by distance from the interacted point. At first, I chose the center of the note's bounding box as the "characteristic point".

However, I didn't think this represented the note very well. If someone were to make an app where users need to click on notes, I expect the users to click on the noteheads. That's where I got the idea to create composite shapes and attach them to renderings.

It offers a slight advantage when an accidental is rendered, but the same could be achieve with boxes. Take a look at how the circle is off-center for notes with accidentals:

image

Ultimately, we don't need this feature for notes. I mostly did it as a proof-of-concept for composite shapes and non-rect shapes, but I think the outcome benefits the user. Let me know if you feel strongly otherwise, modifying the note interaction model shapes is very simple.

jaredjj3:
Openmusicsheetdisplay has hundreds of [EngravingRules](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/blob/aa1472ecb0b4882753676af2acdc41799301200f/src/MusicalScore/Graphical/EngravingRules.ts). As `vexml` improves its customizability, I expect the number of options to grow to this order of magnitude. In this PR, I took a detour to make configuration easier to manage.

Whenever a new configuration key is declared, a control of the corresponding type is automatically added to the vexml.dev playground. This configuration will survive page reloads.

https://github.com/stringsync/vexml/assets/19232300/7a3088cc-c71a-4186-b0af-4de45f15ea29

I'm submitting this PR now so that it's available for people to use. I'll finish adding events in a downstream PR.

rvilarl:
> It offers a slight advantage when an accidental is rendered, but the same could be achieve with boxes. Take a look at how the circle is off-center for notes with accidentals:
> 
> image

Vexflow includes the accidentals, flags and other modifiers in the bounding box. Of course bounding boxes are sometimes too big. See third note.

image

PR #231, opened 2024-06-20
Add more event support

jaredjj3:
This PR fixes #227 by adding a basic event system. It adds an overlay element, which makes the event behavior consistent between SVG and canvas. Without this, SVG elements would "steal" touchend and touchcancel. I decided to leave the event system very bare and simple for now. I'll try to revisit before 1.0.0.

### DIFF
--- a/site/src/components/EventTypeForm.tsx
+++ b/site/src/components/EventTypeForm.tsx
@@ -1,0 +1,49 @@
+import * as vexml from '@/index';
+import { useId, useState } from 'react';
+
+const EVENT_TYPES = ['click', 'hover', 'enter', 'exit'] as const;
+
+export type EventTypeFormProps = {
+  defaultEventTypes: vexml.EventType[];
+  onEventTypesChange: (eventTypes: vexml.EventType[]) => void;
+};
+
+export const EventTypeForm = (props: EventTypeFormProps) => {
+  const [eventTypes, setEventTypes] = useState<vexml.EventType[]>(props.defaultEventTypes);
+
+  const suffix = useId();
+  const id = (eventType: vexml.EventType) => `${eventType}-${suffix}`;
+
+  const onChange = (eventType: vexml.EventType) => (event: React.ChangeEvent<HTMLInputElement>) => {
+    let nextEventTypes: vexml.EventType[];
+    if (event.target.checked) {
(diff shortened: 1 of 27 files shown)


jaredjj3:
> > It offers a slight advantage when an accidental is rendered, but the same could be achieve with boxes. Take a look at how the circle is off-center for notes with accidentals:
> > image
> 
> Vexflow includes the accidentals, flags and other modifiers in the bounding box. Of course bounding boxes are sometimes too big. See third note.
> 
> image

@rvilarl let's continue the conversation from https://github.com/stringsync/vexml/pull/230#issuecomment-2179813848 here.

I'm not getting the same result as the test. I'm using `vexflow.StaveNote.getBoundingBox`.

image

In the tests, I saw [something](https://github.com/0xfe/vexflow/blob/783af8671d6d5c6861ba80505b4cfa7834e30131/tests/stavenote_tests.ts#L104) dealing with `ModifierContext` that I'm not doing in vexml explicitly. I logged `vexflow.StaveNote.getModifierContext` and got this:

image

I suspect that this `ModifierContext` is incorrect. What's the correct way to use this so that the bounding box includes the modifiers?

rvilarl:
To me the ModifierContext looks good: there is one Accidental and one StaveNote as expected.
What version of VexFlow are you using now? Are you adding the Modifiers to the notes?

jaredjj3:
I'm using `vxflw-early-access 5.0.0-alpha.11` and I'm using `vexflow.StaveNote.addModifier` ([code](https://github.com/stringsync/vexml/blob/588b2191c0de1abf0c60c461d58c3743f8f865a2/src/rendering/note.ts#L265)).

rvilarl:
> I'm using `vxflw-early-access 5.0.0-alpha.11` and I'm using `vexflow.StaveNote.addModifier` ([code](https://github.com/stringsync/vexml/blob/588b2191c0de1abf0c60c461d58c3743f8f865a2/src/rendering/note.ts#L265)).

@jaredjj3 please try with `vexflow 5.0.0-alpha.4` we do not release now in vxflw-early-access.

rvilarl:
@jaredjj3 this is why I said that I had been working in BoundigBox. I resolved some problems but I expect more bugs. Let me know if you come across them.

jaredjj3:
@rvilarl will do, thanks for improving it!

PR #232, opened 2024-06-24
Create playback cursor

### LINKED ISSUE #158: Create mechanism to programmatically navigate rendering

jaredjj3:
This PR starts to address #158, but does NOT fix it. I won't have capacity to address this for a while, so I at least wanted to merge some helpful changes now.

> [!NOTE]  
> To self: The last thing I started to work on was a representation of a playback sequence. A playback sequence would be created for each part, and this would be the mechanism by which a discrete or continuous playback cursor could be made. When I get capacity again, I need to finish fleshing out playback/sequence.ts and actually create a sequence, probably in the `Rendering` class.


jaredjj3:
> > I'm using `vxflw-early-access 5.0.0-alpha.11` and I'm using `vexflow.StaveNote.addModifier` ([code](https://github.com/stringsync/vexml/blob/588b2191c0de1abf0c60c461d58c3743f8f865a2/src/rendering/note.ts#L265)).
> 
> @jaredjj3 please try with `vexflow 5.0.0-alpha.4` we do not release now in vxflw-early-access.

RE: https://github.com/stringsync/vexml/pull/231#issuecomment-2185646096

@rvilarl, thanks that worked! I'll remove the circles from the noteheads.

jaredjj3:
It is vexml's test environment. I opened https://github.com/vexflow/vexflow/issues/211.

jaredjj3:
I'm taking a detour in this PR to improve the test environment using vitest. So far, the unit tests are significantly faster than `jest`, and it seems easier to configure the environment (https://vitest.dev/guide/environment#custom-environment). It would be nice to get rid of the transform pipeline needed for jest.

After that, I'll use the environment solution from https://github.com/jsdom/jsdom/issues/3363#issuecomment-1921575184.

jaredjj3:
`vitest` turned out to be less performant than `jest` overall, and it flaked on tests that rely on real canvas behavior. I'm sticking with `jest`.

@rvilarl, the fixes for the tab measure numbers and vexflow modifiers also came through! Thanks again :)

PR #233, opened 2024-09-16
update to vexflow beta

rvilarl:
@jaredjj3 please help me fixing the error.
```
  [RuntimeError] NoStave: No stave attached to instance.

      77 |   @util.memoize()
      78 |   getWidth(): number {
    > 79 |     return this.getVfKeySignature().getWidth() + KEY_SIGNATURE_PADDING;
```

We need to set the stave before calling getWidth but I do not know how to get access to the stave in keysignature.ts

```
/** Returns the width of the key signature. */
  @util.memoize()
  getWidth(): number {
    return this.getVfKeySignature().getWidth() + KEY_SIGNATURE_PADDING;
  }
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   },
   "dependencies": {
     "jszip": "3.10.1",
-    "vexflow": "5.0.0-alpha.4"
+    "vexflow": "^5.0.0-beta.0"
   },
   "devDependencies": {
     "@babel/core": "^7.17.8",
--- a/yarn.lock
+++ b/yarn.lock
@@ -6328,10 +6328,10 @@ vexflow-fonts@^1.0.6:
   resolved "https://registry.yarnpkg.com/vexflow-fonts/-/vexflow-fonts-1.0.6.tgz#2fe3e3a056850bb39f436c568e50dec2c95ab90c"
   integrity sha512-B6jJGfCFn0OJoC4UdGq4FCa1tEtIruaBwyb/BPAeqyROr+gTJ4FlcM4X0ir5xdoH2CUrRwNoHdUWVIeqqIEj4g==
 
-vexflow@5.0.0-alpha.4:
-  version "5.0.0-alpha.4"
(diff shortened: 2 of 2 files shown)


jaredjj3:
I'm on mobile, apologies for the lack of formatting. Hopefully this unblocks you.

Right now, vexflow.Stave objects are created downstream in rendering.Part.getStaves. Some possible options:

- Create a temporary vexflow.Stave solely for measuring width. 
- Create vexflow.Stave objects in rendering.StaveSignature and pass it accordingly to rendering.KeySignature.

rvilarl:
Thanks! I am considering if it makes sense to have the stave. This is required to calculate the bounding box position correctly but not the height and width.

jaredjj3:
In keysignature.ts, this seemed to fix it:

```ts
  private getVfKeySignature(): vexflow.KeySignature {
    const vfStave = new vexflow.Stave(0, 0, 0);
    const vfKeySignature = new vexflow.KeySignature(
      this.getKey(),
      this.previousKeySignature?.getKey() ?? undefined,
      this.getAlterations()
    ).setPosition(vexflow.StaveModifierPosition.BEGIN);
    vfKeySignature.addToStave(vfStave);
    return vfKeySignature;
  }
```

However, I've been having issues running the test via `yarn test` (Docker) and `yarn jest`. The test setup probably needs to be updated.

jaredjj3:
I was able to run the tests. The change I mentioned does produce diffs, but they're negligible.

rvilarl:
I have set a PR in vexflow to avoid changes here.

jaredjj3:
Thanks! I think that's a better solution since it will prevent other vexflow users from breaking.

jaredjj3:
@rvilarl, I'm preparing to have this ready for VexFlow 5's release. Feel free to reach out to me when the issue we talked about has been addressed in VexFlow.

jaredjj3:
@rvilarl, just FYI, I did this work in https://github.com/stringsync/vexml/pull/256.

rvilarl:
@ronyeh could you release a beta?

rvilarl:
@jaredjj3 this should be fixed in the vexflow 5.0.0-beta.1
your work around code should not be needed.

jaredjj3:
@rvilarl removed in #275. Thanks!

PR #239, opened 2024-12-11
Implement smooth interpolated seeking

jaredjj3:
This PR implements `Cursor.seek`, which is the interpolated alternative to `Cursor.snap`.

https://github.com/user-attachments/assets/e56e163f-a625-48cc-816b-a99b6e735bb0

It also supports repeats!

https://github.com/user-attachments/assets/1c41c1c9-2dc8-4903-af03-7b8242447f9a

Next, I think I'll introduce a max height for the vexml container, and allow the cursor to scroll the container if the cursor is not in view. This will be an option that's controlled in `Cursor.seek` or `Cursor.snap`.


xdavidliu:
nice

PR #247, opened 2024-12-18
Add w3c MusicXML test suite

### LINKED ISSUE #163: Official MusicXML examples
I thought I'd let you know that I wrote a script to scrape the [MusicXML examples from the W3C site](https://www.w3.org/2021/06/musicxml40/musicxml-reference/examples/).

Here's [the script](https://github.com/infojunkie/musicxml-midi/blob/main/src/js/musicxml-examples.js) and here are [the saved examples](https://github.com/infojunkie/musicxml-midi/tree/main/test/data/examples). 

Unfortunately, most of those are incomplete MusicXML fragments that lack the overall "score" or "measure" structures. I would be willing to modify the scraper to add those structures, if you are interested to incorporate them into your tests.

jaredjj3:
This PR fixes #163 by adding the MusicXML test suite from [infojunkie/musicxml-midi](https://github.com/infojunkie/musicxml-midi). All but one test is commented out. If/when I strive for better coverage, I will uncomment the rest and evaluate their correctness.


jaredjj3:
It's worth mentioning that the examples are readily available on the dev site (https://vexml.dev).

PR #266, opened 2025-02-23
Render Document inputs

### LINKED ISSUE #265: Interface for rendering a parsed `XMLDocument`?
Hi,

First of all - amazing project, it opens up so many possibilities for building interesting music apps!

I have a use-case where I am programatically generating MusicXML in JS. Currently I have to do a useless serialization<->deserialization roundtrip to use Vexml, as such:

```
const musicXML: XMLDocument = ...
const parser = new vexml.MusicXMLParser(...);
const document = parser.parse(
    new XMLSerializer().serializeToString(musicXML),
);
```

It would be great to have an interface for directly passing the `XMLDocument`. My understanding is that there is no such way currently. Happy to submit a PR if you think this is useful.

jaredjj3:
This PR fixes #265 by extending the `parsing.MusicXMLParser` and `renderMusicXML` signatures to accept `XMLDocument` objects.


jaredjj3:
@yfyf, I used `Document` instead of `XMLDocument` because `XMLDocument extends Document`. Your use case should be covered. This PR will be included in ^0.1.2.

Issue #260, opened 2025-02-15
Pianoplay integration


rvilarl:
@jaredjj3 I have created this issue to get your assistance on the integration with Pianoplay.

rvilarl:
1. you have not published vexml to npm, right? I tried to install it with `npm install vexml`

rvilarl:
> 1. you have not published vexml to npm, right? I tried to install it with `npm install vexml`

Resolved reading documentation `npm install @stringsync/vexml` sorry

rvilarl:
First version, score rendered with vexml and vexflow 5!
@jaredjj3 this is so exciting!

https://github.com/user-attachments/assets/f7312b44-0246-475a-98b9-0f4fe4bf3cc5

jaredjj3:
@rvilarl wow, that looks really good! Did you change any of the default options? Were there any options you wish you had?

rvilarl:
Now I have to integrate the cursor. How do I get the notes under the cursor in order to play them? I will push the current running version in a vexml branch in pianoplay.

rvilarl:
> [@rvilarl](https://github.com/rvilarl) wow, that looks really good! Did you change any of the default options? Were there any options you wish you had?

This is just the vanilla call. It worked out of the box!

jaredjj3:
>How do I get the notes under the cursor in order to play them?

I'm glad you brought this up. I'm really curious how well this works for your use case. I was really unsure about the API when developing this.

1. Follow these instructions to create a cursor model and component: https://github.com/stringsync/vexml/blob/c4511e156e317704577c07793fe87828be1df448/src/playback/types.ts#L7. Let me know if anything is wrong with the instructions and I'll fix it.

2. On the cursor **model**:

```ts
cursor.addEventListener('change', (state) => {
  console.log(state);
});
```

You'll see `anchorElement` (the element that triggered the change) and `activeElements` (the elements that are currently being played, which I think includes the `anchorElement`) properties. From there, you can type cast the elements by checking the `Element.name`, and grab the pitch and octave data you're interested in.

Let me know how it goes!

rvilarl:
Here it is!

![Image](https://github.com/user-attachments/assets/80a65b00-30ad-417c-a57c-f2aa2b4692d9)

*Two issues*
- the cursor does not cover the two instruments
- there seems to be only one active element

*Two questions*
- how do I move the cursor to the next position?
- is it possible to have more than one cursor?

*Feedback to README*
[...]

PR #280, opened 2025-04-11
Move CursorFrameHint to CursorStateHint

jaredjj3:
This PR fixes the bug mentioned in https://github.com/stringsync/vexml/issues/260#issuecomment-2791534444. It will be released in 0.1.7.

- Moves `CursorFrameHint` to `CursorStateHint`.
- Adds `StartHint` and `StopHint`.
- Keeps `CursorStateHint[]` calculation lazy.
- Adds light test coverage to `CursorStateHint` (checks that getting hints does not throw).
- Improves playback tests by making each test more configurable.
- Adds `Cursor.iterable`.
- Fixes `rendering.Document` bug described in https://github.com/stringsync/vexml/issues/260#issuecomment-2796639572 by allowing `elements.Note` to return multiple pitches.
- Surfaces Note subtypes in `elements.Note.getSubtype` (note or chord).

I added _playback_chords.musicxml_ to the vexml playback test suite, which is a snippet from https://github.com/stringsync/vexml/issues/260#issuecomment-2791534444.

image


rvilarl:
@jaredjj3 where are the hints now? Sorry I do not find them.

jaredjj3:
cursor.getCurrentState().hints.get()

rvilarl:
I get

Property 'getCurrentState' does not exist on type 'CursorState'.ts(2339)

jaredjj3:
If you have a CursorState object, it should just be cursorState.hints.get().

rvilarl:
Hi @jaredjj3 

I have two doubts:
1. If only one pitch of a chord is tied to another note, how is this going to be presented?
2. I would expect to get start first and then stop, but the first hints in the score come with stop.

image

jaredjj3:
>1. If only one pitch of a chord is tied to another note, how is this going to be presented?

All the pitches existing in the previous state will also have a sustain hint, even though that's not completely accurate. If this is undesirable, would you file a bug please?

That's a shortcoming of the way `elements.Note.sharesACurveWith` implemented:

https://github.com/stringsync/vexml/blob/1bfc4f7dc7cdb0b47abb25095026bfd54a7d4f10/src/elements/note.ts#L53-L55

I think I need to create an `elements.ChordNote` and implement an `elements.ChordNote.sharesACurveWith(chordNote: ChordNote)`.

>2. I would expect to get start first and then stop, but the first hints in the score come with stop.

Me too, that's definitely a bug. Right now, the [tests](https://github.com/stringsync/vexml/blob/1bfc4f7dc7cdb0b47abb25095026bfd54a7d4f10/tests/unit/playback/lazycursorstatehintprovider.test.ts) just check that nothing is thrown when getting hints. However, I think this warrants more thorough tests.

rvilarl:
@jaredjj3 I have a suggestion. What about making the hints pitch based rather than note based? Then we would have the types start, stop, keep.

jaredjj3:
The reason it's based on notes is to support more use cases. For example, if you want to color and uncolor the notes based on if they're active or not, just having the pitch is not enough information.

PR #283, opened 2025-05-01
Fix hints and improve unit tests

### LINKED ISSUE #281: Improve LazyCursorStateHintProvider unit tests
They're returning incorrect data, see: https://github.com/stringsync/vexml/pull/280#issuecomment-2799197760

jaredjj3:
This PR fixes #281. It will be released in 0.1.8.

- This PR formalizes the describer pattern and uses it in existing Timeline and CursorFrame tests.
- Marks MeasureSequenceIterator as LegacyMeasureSequenceIterator to communicate intent of eventually replacing it.


jaredjj3:
@rvilarl, I'm going to release a vexml patch with this PR so that you're unblocked. There may be still issues with repeats and chords.

